PR #7, opened 2017-09-24
Add a new createHttpClient() method to NetworkUtilities

melissalinkert:
createHttpClient() constructs a DefaultHttpClient with no URL and a 30
second timeout.  All instantiations of DefaultHttpClient with this
timeout are also updated to use the new method.

This commit should not change any existing behavior; the only purpose is
to reduce duplicated code.


deepwinter:
Looks good, thanks for your work!

PR #14, opened 2017-11-06
Add two API, trees/details/user, trees/create

lastforeverzl:
Add API module. Created a DataManager class to manage HTTP response, abstract methods should be implemented in fragment to make UI change. "trees/details/user" works fine, but create a new tree api return 405 error code. @deepwinter


deepwinter:
OK, verified 405 error code when using curl as well.  I will have to check on the server side to isolate the issue.

PR #24, opened 2017-12-12
Change to dospaces cloud storage

lastforeverzl:
Upload tree images to Digital Ocean Spaces storage. Return the image url to the server rather than base 64 image format. This function will compatible with the new server infrastructure, so keep code in this branch until new server API release. Access key and secret key should be changed in the gradle.properties file and keep it local.


deepwinter:
Merged this one manually.

Issue #42, opened 2018-02-13
Android deployment automation


Davidezrajay:

mckornfield:
Do you mind providing a bit more context on what this means? Not super familiar with Android but trying to get a sense of what this entails.

deepwinter:
At one point in time, there were services where I could push a tag to a specific branch on github, and the service/server would build the app and send it out to fabric.io for deployment to our testers.  This makes it more straightforward to deploy nightly builds.  It would also be good to be able to do this when sending to the play store.  There may be options that just use gradle and do this locally without a server, or there may be build server options.  First step is to assess what how people are currently addressing this need in devops, last time I set something like this up was a couple years ago.

Davidezrajay:
https://github.com/codepath/android_guides/wiki/Automating-Publishing-to-the-Play-Store

Davidezrajay:
https://github.com/marketplace/category/continuous-integration

deepwinter:
We made progress on this elsewhere.

PR #64, opened 2018-03-18
Initial Gradle plugin for Appstore Deployment

mckornfield:
Part of Issue #42 
Still requires hooking into some sort of CD mechanism, but at least allows doing something from the command line.

### DIFF
--- a/README.md
+++ b/README.md
@@ -5,6 +5,18 @@ This project coordinates tree planting employment for people living in extreme p
 The Android segment allows people to track and verify reforestation plantings,
 paying planters on a per planting basis.
 
+## Deployment
+
+There is one prerequisite to using the appropriate gradle tasks:
+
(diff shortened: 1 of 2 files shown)


mckornfield:
@deepwinter This was my initial crack at it... I followed 
https://github.com/codepath/android_guides/wiki/Automating-Publishing-to-the-Play-Store
and the github repo which is a bit more up to date
https://github.com/Triple-T/gradle-play-publisher

I didn't do the google service account portion because I wasn't sure how you guys had that set up, but this can at least be a starting point... I think this would push the files to beta and then you could manually move to prod in the PlayStore window... but I am not that familiar with it myself.

I can also do a readme update as part of this... I'll at least put the relevant gradle tasks.

deepwinter:
Cool.  I will check this out a bit soon.

PR #82, opened 2018-04-19
Solution 55

beastlyCoder:
Instead of having the app crash like in issue number 55 which was due to the location accuracy not being found I displayed a toast that said that the location accuracy was not being set instead of crashing the whole application.

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.java
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.java
@@ -164,11 +164,8 @@ public class CameraActivity extends Activity implements PictureCallback, OnClick
 			e.printStackTrace();
 		}
 
-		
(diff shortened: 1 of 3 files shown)


Davidezrajay:
@deepwinter have you had a chance to look at this one?

deepwinter:
This is a decent fix, however from the stack trace the problem is not the accuracy isn't found, but rather that the location itself is null.

If you look below the try/catch you can see that the code asks again for getAccuracy and will crash there instead because mCurrentLocation is still null.

```
contentValues.put("accuracy",
 				Float.toString(MainActivity.mCurrentLocation.getAccuracy()));	 				
```

@beastlyCoder so what has to happen is if mCurrentLocation is null, we have to disallow the user from creating a new record.  And furthermore, we should determine how they are getting to this point in the app without a location set (they should not be able to click the + button and get this far into the process of creating a tree).

PR #94, opened 2018-05-10
Fixed issue 90

### LINKED ISSUE #90: Update About Page Text
Please add the following text to the about page.

"This tree tracking planter app is made to track planted trees while supporting people and organizations that plant, monitor and protect trees. 

Take a picture of your trees and then press sync and your tree data will be uploaded to a map at www.treetracker.org. 

Visit www.GreenStand.org for more details on how this system works.

Your app requires 10 meters of accuracy to take a photo. 

Email: info@greenstand.com for more access to your data, or a web link with your specific trees. 

This app is under development by Zaven Arra and his cool team of volunteers at GreenStand.org. 

Developers can find the open source project code at www.github.com/greenstand"

beastlyCoder:
Fixed the text in the about 

fix #90

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -15,7 +15,7 @@ apply plugin: 'com.github.triplet.play'
 
 android {
     compileSdkVersion 25
-    buildToolsVersion "26.0.2"
+    buildToolsVersion '27.0.3'
     playAccountConfigs {
         defaultAccountConfig {
             jsonFile = file('keys.json')
--- a/app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.java
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.java
@@ -205,6 +205,7 @@ public class CameraActivity extends Activity implements PictureCallback, OnClick
 	    } else {
 	        return null;
 	    }
+
 	    return mediaFile;
 	}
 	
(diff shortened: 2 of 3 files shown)


Davidezrajay:
Looks like a fix for #98 

"Zaven noted that closing the database was not necessary
    `public synchronized void closeDatabase() {`	
Further looking at the code there is no uses of this method
        `mOpenCounter--;	
//    public synchronized void closeDatabase() {
        if(mOpenCounter == 0) {	
//        mOpenCounter--;
            // Closing database	
//        if(mOpenCounter == 0) {
            mDatabase.close();	
//           
 // Closing database
        }	
//            mDatabase.close();
    }	//        }
//    }
`

ryan-beckett:
I already commented on this in Slack. That code is buggy for one. Secondly, we really don't need to close the DB ever.

deepwinter:
This PR is suspect - there are several places where there are symmetrical openDatabase/closeDatabase calls, so removing the function doesn't make sense.  I think this code was originally written from the activity vantage point, not wanting to have a notion of a global 'Application'.  Still doesn't make great sense though, because then you would close the database in each activity, not in a static application-wide manager.  I think we may have database related crashes due to locking between background and foreground threads   see: https://stackoverflow.com/questions/4557154/android-sqlite-db-when-to-close

I am rejecting this PR.

deepwinter:
This PR is rejected.

Issue #98, opened 2018-05-20
DatabaseManager - org.greenstand.android.TreeTracker.database.DatabaseManager.openDatabase(2 crashes)


Surbhi79:
Fatal Exception: java.lang.NullPointerException: Attempt to invoke virtual method 'android.database.sqlite.SQLiteDatabase org.greenstand.android.TreeTracker.database.DbHelper.getWritableDatabase()' on a null object reference
       at org.greenstand.android.TreeTracker.database.DatabaseManager.openDatabase(Unknown Source)
       at org.greenstand.android.TreeTracker.fragments.DataFragment.updateData(Unknown Source)
       at org.greenstand.android.TreeTracker.fragments.DataFragment.onResume(Unknown Source)
       at android.support.v4.app.Fragment.performResume(Unknown Source)
       at android.support.v4.app.FragmentManagerImpl.modifiesAlpha(Unknown Source)
       at android.support.v4.app.FragmentManagerImpl.moveFragmentToExpectedState(Unknown Source)
       at android.support.v4.app.FragmentManagerImpl.modifiesAlpha(Unknown Source)
       at android.support.v4.app.BackStackRecord.addOp(Unknown Source)
       at android.support.v4.app.FragmentManagerImpl.executeOps(Unknown Source)
       at android.support.v4.app.FragmentManagerImpl.executeOpsTogether(Unknown Source)
       at android.support.v4.app.FragmentManagerImpl.optimizeAndExecuteOps(Unknown Source)
       at android.support.v4.app.FragmentManagerImpl.execPendingActions(Unknown Source)
       at android.support.v4.app.FragmentManagerImpl$1.run(Unknown Source)
       at android.os.Handler.handleCallback(Handler.java:836)
       at android.os.Handler.dispatchMessage(Handler.java:103)
       at android.os.Looper.loop(Looper.java:203)
       at android.app.ActivityThread.main(ActivityThread.java:6251)
       at java.lang.reflect.Method.invoke(Method.java)
       at com.android.internal.os.ZygoteInit$MethodAndArgsCaller.run(ZygoteInit.java:1073)
       at com.android.internal.os.ZygoteInit.main(ZygoteInit.java:934)

Surbhi79:
[...]

PR #102, opened 2018-06-13
Possible solution for #98

Surbhi79:


ryan-beckett:
I don't think the change here solves the issue. Consider the following code:

```
public DbHelper(Context context, String name, CursorFactory factory, int version) {
        super(context, DB_NAME, null, 1);
        this.myContext = context;
}
```

_name_ isn't even used.

Issue #109, opened 2018-07-19
Buggy Password Confirmation for Sign Up Page


ryan-beckett:
When a new user signs up, even when their passwords match, an error message is displayed stating "The passwords do not match". This is due to a minor code bug.

deepwinter:
Looks like the PR for this issue was merged, closing ticket.

PR #110, opened 2018-07-19
Fix for "Buggy Password Confirmation for Sign Up Page" #109

ryan-beckett:
The confirmation is confirmed working now. I must not have tested my first pull request. I also removed the hide/show icon since the `TextInputEditText` LAF differed from `EditText`, and to me just gave an inconsistent feel.


deepwinter:
Thanks for the fix!

Issue #112, opened 2018-07-24
Memory Optimisation Code


dezidaisy:
https://medium.freecodecamp.org/how-we-reduced-memory-footprint-by-50-in-our-android-app-49efa5c93ad8

deepwinter:
@desaisameera See my comments in the PR for this, I'll need to see some memory profiling results before merging this work.

Elforama:
@deepwinter Is this issue still needed?

Davidezrajay:
This one seems to have been fixed on the move to Kotlin. To be evaluated again later

PR #113, opened 2018-07-24
Issue 112

dezidaisy:
Fix for Issue 112

### DIFF
--- a/app/fabric.properties
+++ b/app/fabric.properties
@@ -1,0 +1,3 @@
+#Contains API Secret used to validate your application. Commit to internal source control; avoid making secret public.
+#Tue Jul 24 22:38:50 BST 2018
+apiSecret=89e38cb893538f0a2d0c511ae9ee802e1aa0803357faf1bb0ac60dc21caa3698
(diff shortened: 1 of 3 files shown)


ryan-beckett:
I would have expected to see the use of `BitmapFactory.Options.inBitmap` to implement the re-use of allocated memory.

dezidaisy:
I'm not sure but I think that using BitmapFactory.Options.inBitmap would prevent reserving memory used to declare it early on but it has its own limitations like size issues, etc..

ryan-beckett:
I don't follow. The docs state that `inBitMap` when not null will used to hold the bitmap, so new heap space isn't allocated. I believe the "same-size" limitation was prior to API 19. I'm not trying to be combative, I just don't see why that field isn't set, especially when the tutorial you referenced did exactly that.

dezidaisy:
Yes but we may not need those specifications. They might be restrictive. I think this simple approach will improve things. Anyway, no harm done!

deepwinter:
Has this fix been tested in our app?  Before I merge this I would need to see some before/after memory allocation profiling, like in the article referenced in #112

dezidaisy:
I will test it out soon.

dezidaisy:
Sorry not able to test it out. I keep getting insufficient accuracy.

deepwinter:
You should build against the build variant 'dev' instead of 'debug'.  The dev build variant allows for creating trees without a specific accuracy.

dezidaisy:
Ok I will try that.. :)

Issue #117, opened 2018-08-18
Tree Preview shows 0 meter accuracy for all trees


deepwinter:
Should show recorded accuracy.

deepwinter:
Fixed in 7b02d15a2a6c5cdfe6d8d8d75b75e66ee57e4423

PR #129, opened 2018-09-01
first batch of refactoring to Kotlin and replacing AsyncTasks

Anup0101:


Davidezrajay:
@Anup0101 Super awesome... It will be reviewed shortly

deepwinter:
Hi Anup, please chat me on Slack.  I was looking over your code and we should discuss the move to rxjava vs use coroutines.  I had started in on the switch to coroutines, but could be convinced to use rxjava.

deepwinter:
@Anup0101 Can you please submit a separate pull request for issue #117 ?

Anup0101:
@deepwinter No problem, I'll have a separate small pr just for the accuracy soon. RxJava in my opinion can definitely be written in a sequential manner, chaining as many operations as you'd like to. Everything is treated as a stream, yes, but that shouldn't affect performance the way it's used here, but yeah I'll reach out about the same in a bit.

Anup0101:
There weren't any merge conflicts this morning, think it's just behind today's changes. Anyway no problem man, this PR and Rx is just a suggestion from I was working on but if you feel coroutines are the way to go that's cool too. I'll have a PR for those 2 changes and let's talk tomorrow evening, works much better for me, about Rx vs coroutines. I personally am for Rx but I agree that it may add GC overhead. Still, Rx is very useful.

deepwinter:
@Anup0101 Sounds great.  I'm not against Rx, it's just that there were already some coroutines in place, and I'd like to make the right choice and not have both in the codebase.  I'm not very familiar with Rx, but I'm definitely open to the suggestion, as I've just recently become familiar with coroutines and kotlin anyway.

Anup0101:
Yeah, I read through the SyncTask replacement with coroutines instead of Rx. I agree with you about a performance benefit, it's just that given the minor difference(from what I gather) I figured Rx would continue to prove to be more useful as the app grows and usually makes reactive code well structured imo. I've put up a new pr with the couple of old changes.

deepwinter:
We have incorporated this work elsewhere, and this PR won't be merged.  Closing this one out.  Thanks for your work!!

PR #154, opened 2018-10-01
More camera changes

Anup0101:


Anup0101:
Since this hasn’t been tested yet, I’d like to point out that you just need to choose a primary vs secondary camera instance and assign the width/height you want for the captured image. All else for the camera2 implementation should be good here.

Davidezrajay:
@Anup0101 thanks for this. Can you test it and make sure it works?

Anup0101:
No prob, I can look into it this weekend.

app/build.gradle
```diff
@@ -38,7 +38,7 @@ android {
     }
     defaultConfig {
         applicationId "org.greenstand.android.TreeTracker"
-        minSdkVersion 15
+        minSdkVersion 21
```
Elforama: This app is for 3rd world countries. Do we really want to ignore users with old phones?

Elforama:
I'm going to close this PR. Updates to the camera are now being worked on using the CameraX jetpack library.

PR #184, opened 2018-11-24
Discuss: Map page

Peacenux:


Davidezrajay:
Thanks, @Peacenux, here are my thoughts

"Add Tree" should be stand out. Planters are pushing this button thousands of times a day. 

There should only be one button in the place of the 'Sync and Stop' buttons. When pressing "sync," the button should change to "Stop"

PR #193, opened 2018-11-28
chore: Updated libraries, Gradle and Kotlin versions

LaviniaDragunoi:
The update of the libraries produced more changes inside activities and
fragments.
In project build.gradle I declared the maps dependency separated because
is creating errors when was generally declared. For more info see:
 https://developers.google.com/android/guides/setup#split


deepwinter:
Great, thanks!

PR #209, opened 2019-01-15
update: Migrate to AndroidX

LaviniaDragunoi:
Thanks!


LaviniaDragunoi:
The last commit was to replace all the findVieById with Kotlin Android extensions feature, this will make the app to use findViewById only once for each view and save in a cache the result for the future usage. 
Thanks!

PR #215, opened 2019-01-23
Update depricated

LaviniaDragunoi:
Updated libraries and eliminated some warnings.
Thanks

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.kt
@@ -10,7 +10,6 @@ import android.graphics.Bitmap
 import android.graphics.BitmapFactory
 import android.graphics.Matrix
 import android.hardware.Camera
-import android.media.ExifInterface
 import android.net.Uri
 import android.os.Bundle
 import android.os.Environment
@@ -18,6 +17,7 @@ import androidx.core.app.ActivityCompat
 import android.view.View
 import android.widget.ImageButton
 import android.widget.ImageView
+import androidx.exifinterface.media.ExifInterface
 import kotlinx.android.synthetic.main.camera_preview.*
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.android.UI
@@ -34,7 +34,8 @@ import java.text.SimpleDateFormat
 import java.util.*
 
 
-class CameraActivity : Activity(), Camera.PictureCallback, View.OnClickListener, ActivityCompat.OnRequestPermissionsResultCallback {
+class CameraActivity : Activity(), Camera.PictureCallback, View.OnClickListener,
+    ActivityCompat.OnRequestPermissionsResultCallback {
 
     private var mCamera: Camera? = null
     private var mPreview: CameraPreview? = null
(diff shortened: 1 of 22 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.kt
```diff
@@ -60,7 +61,7 @@ class CameraActivity : Activity(), Camera.PictureCallback, View.OnClickListener,
         captureButton!!.setOnClickListener(this@CameraActivity)
 
         if(intent.extras != null) {
```
mozarty: better to use intent.extras?.let {

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
```diff
@@ -73,10 +73,10 @@ class MainActivity : AppCompatActivity(), ActivityCompat.OnRequestPermissionsRes
         if (mSharedPreferences!!.getBoolean(ValueHelper.FIRST_RUN, true)) {
```
mozarty: if mSharedPreferences could be null here it is better to do a null check instead of !!

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
```diff
@@ -149,7 +150,8 @@ class MainActivity : AppCompatActivity(), ActivityCompat.OnRequestPermissionsRes
 
             fragmentTransaction = supportFragmentManager
                     .beginTransaction()
-            fragmentTransaction!!.replace(R.id.containerFragment, fragment as DataFragment).addToBackStack(ValueHelper.DATA_FRAGMENT).commit()
+            fragmentTransaction!!.replace(R.id.containerFragment, fragment as DataFragment)
```
mozarty: please use `?` instead of `!!`

app/src/main/java/org/greenstand/android/TreeTracker/database/DbHelper.kt
```diff
@@ -80,9 +82,9 @@ class DbHelper(private val myContext: Context, name: String, factory: CursorFact
                 while (treesCursor.moveToNext()) {
 
 
-                    val sharedPreferences = myContext.getSharedPreferences(
+                    val sharedPreferences1 = myContext.getSharedPreferences(
```
mozarty: i don't understand the rename , if there is another variable with the same name please use a different meaningful name , i.e. userSharedPrefrences

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
LaviniaDragunoi: On 73 line (MainActivity) did you said that I should replace the line `if (mSharedPreferences!!.getBoolean(ValueHelper.FIRST_RUN, true)) {` with `if (mSharedPreferences**?.**getBoolean(ValueHelper.FIRST_RUN, true)) {` Because When I do this inside the if statement is a Type mismatch, we will have a Boolean? instead of a Boolean type that is required.

PR #233, opened 2019-03-16
226 tree height

Elforama:
This adds the Tree height feature to the New Tree flow. 
The New Height Feature is hidden behind `FeatureFlags.TREE_HEIGHT_FEATURE_ENABLED`. This flag is set to true atm. It still needs a flavor to switch this flag on/off

Things to note:

- Moves a lot of DB operations into the TreeManager
- Adds some logic to modify the New Tree screen based on the feature flag.
- Added Kotlin serialization by enabling experimental flag (this is to allow easy passing of objects through Bundles)
- Added Googles life cycle libraries which is mainly for the ViewModel.
- TreeHeightFragment follows MVVM design pattern using LiveData objects to allow view to listen to events from ViewModel class.


deepwinter:
Top notch!    It's rad to have this work getting done so quickly.

There are are just two minor things that I'd like to address before merging the PR

1) I don't think we want to remove the 'notes' feature when tree height is enabled.   Was this removed to clean up the UI?  I think it would make sense to go to the tree heigh color picker directly after taking the photo, and then show the tree image + note entry view with 'save' before returning to the map.

2) I had originally imagined the 'attributes' table just as a treeid-key-value dump table of any attributes we'd like to tag trees with not or in the future.  You implemented it more as a traditional relational database table with named columns.  I am imagining that various app flavors will, over time, have a variety of custom attributes, and we'd like to support them with the least work possible, since volunteers come and go.  What do you think about reorganizing this into key value style table, such that each attribute has its own row for a given tree?  The main concern I have here is easily adding custom attributes without needing database migrations and/or created a large tree_attributes table with many empty columns.  Open to other ideas as well.

Elforama:
@deepwinter 
1. All UX in this PR is per design. Perhaps we should talk with AnnaEye about this.
2. If we want a simple key/values table it might be better to just use the Android SharedPreferences. SharedPreferences can have multiple key/value stores. Each store could be fetch with a treeId, with the values in it being key/value attributes. How does that sound?
e.g. `context.getSharedPreferences(treeId).getInt(treeColorKey)`

deepwinter:
I have very mixed feelings about storing this kind of data in SharedPreferences.  I am happy with the key/value attributes approach being implemented server side, and with the ability for Android send those attributes along, so we can table the client side discussion for the time being and revisit if there ends up being an engineering reason to implement a more nuanced approach.

deepwinter:
I am going to merge this PR and file a ticket against splitting out the deactivation of the notes feature into its own config var.

PR #238, opened 2019-03-25
New login feat

LaviniaDragunoi:


app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+import org.greenstand.android.TreeTracker.database.entity.PlanterDetailsEntity
+import org.greenstand.android.TreeTracker.utilities.ImageUtils
+import org.greenstand.android.TreeTracker.utilities.ValueHelper
+import timber.log.Timber
+
+class LoginFragment : Fragment(){
+    var phoneNumberEntered: String? = null
+    var emailEntered: String? = null
+    private var mPhotoPath: String? = null
+    private var planterId: Int = -1
```
Elforama: If we want to express a "no value" value, we can make this a nullable property

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+
+    override fun onCreateView(inflater: LayoutInflater, container: ViewGroup?, savedInstanceState: Bundle?): View? {
+        val v = inflater.inflate(R.layout.fragment_login, container, false)
+        activity?.toolbarTitle?.setText(R.string.greenstand_welcome_text)
```
Elforama: We can use `requireActivity()` to avoid using `?` everywhere

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+
+        v.sign_up_button?.visibility = View.INVISIBLE
+//Here we will watch the phone edit text to check if the input number(#phoneNumberInput) is in our database already.
+    //if the #phoneNumberInput is in our database we will turn Login button to ActiveButtonStyle and we will set
+    // ClickListener on this button to go on further with the login feature
+     v.phoneEditTextLogin.addTextChangedListener(object: TextWatcher {
+        @SuppressLint("NewApi")
+        override fun afterTextChanged(p0: Editable?) {
```
Elforama: Can we please name `p0` something more meaningful?

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+
+                    override fun beforeTextChanged(p0: CharSequence?, p1: Int, p2: Int, p3: Int){
+                        inactivateLoginButton()
+                    }
+
+                    override fun onTextChanged(p0: CharSequence?, p1: Int, p2: Int, p3: Int) {
+                        inactivateLoginButton()
+                    }
+                })
+                when (isExistingUser(phoneNumberEntered!!)) {
```
Elforama: This `when` logic is duplicated above us on line 62. To reduce repeating ourselves we can move this logic into a function and pass in any parameters we need such as a phone number or email.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+            inactivateLoginButton()
+        }
+
+        override fun onTextChanged(p0: CharSequence?, p1: Int, p2: Int, p3: Int) {
+            inactivateLoginButton()
+        }
+
+    })
+
+    v.sign_up_button.setOnClickListener(object : View.OnClickListener{
```
Elforama: There is a shorthand for this in Kotlin
`v.sign_up_button.setOnClickListener {  your code in here }`

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+
+
+//This is the method that will check into DB if the textEntered(that could be the phoneNumber or the email) already
+// exists in our DB. If exists it will return TRUE else it will return FALSE. in fact, it will return the if the
+// planterId is different than -1 ( this is the initial value given to var planterId), this var will receive
+// the planterId from the DB if the user already exists
+private fun isExistingUser(textEntered: String): Boolean {
+    runBlocking {
+        planterId = GlobalScope.async {
+            TreeTrackerApplication.getAppDatabase().planterDao().getPlanterIDByIdentifier(textEntered)
```
Elforama: Can we put this database access logic into the TreeManager? It can be another method in there.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+
+}
+
+
+@SuppressLint("NewApi")
+fun activateLoginButton(){
+    login_button.setTextAppearance(R.style.ActiveButtonStyle)
+    login_button.setBackgroundResource(R.drawable.button_active)
+    if(sign_up_button.visibility == View.VISIBLE) sign_up_button?.visibility = View.INVISIBLE
+    login_button.setOnClickListener(object : View.OnClickListener{
```
Elforama: We can update this click listener like we did above. Anywhere else you use a click listener should just be
`myView.setOnClickListener { do something here }`

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+
+    })
+
+
+
+}
+
+@SuppressLint("NewApi")
+fun inactivateLoginButton(){
+    login_button.setTextAppearance(R.style.InactiveButtonStyle)
```
Elforama: This works fine. Just want to make you aware of some Kotlin features.
If you want to reduce the amount of times you type login_button.xxx you can use the `with` or `apply` extension functions.
For example, this could become this:
` login_button.apply {  
setTextAppearance(myStyle) 
setOnClickListener(null)
etc...
}

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+ return v
+}
+
+
+//This is the method that will check into DB if the textEntered(that could be the phoneNumber or the email) already
+// exists in our DB. If exists it will return TRUE else it will return FALSE. in fact, it will return the if the
+// planterId is different than -1 ( this is the initial value given to var planterId), this var will receive
+// the planterId from the DB if the user already exists
+private fun isExistingUser(textEntered: String): Boolean {
+    runBlocking {
```
Elforama: This function is blocking the main thread, this means any UI has to wait for this to finish. In this case it's not too bad, but in general we should not be doing this.
One way to resolve this to to have the area this is called, call it inside of a `launch` function. This function can than be made a suspending function, and the runBlocking and async can be removed.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,200 @@
...
+import org.greenstand.android.TreeTracker.activities.CameraActivity
+import org.greenstand.android.TreeTracker.application.Permissions
+import org.greenstand.android.TreeTracker.application.TreeTrackerApplication
+import org.greenstand.android.TreeTracker.database.entity.PlanterDetailsEntity
+import org.greenstand.android.TreeTracker.utilities.ImageUtils
+import org.greenstand.android.TreeTracker.utilities.ValueHelper
+import timber.log.Timber
+
+class LoginFragment : Fragment(){
+    var phoneNumberEntered: String? = null
```
Elforama: Can we make these two properties private?

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
LaviniaDragunoi: It seems that I misunderstood at first the usage of runBlocking. I will read more about this and about launch too. Thanks!

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
LaviniaDragunoi: Can you give more details, please?

app/src/main/java/org/greenstand/android/TreeTracker/database/dao/PlanterDao.kt
```diff
@@ -21,6 +21,10 @@ interface PlanterDao {
     @Query("SELECT * FROM planter_identifications WHERE photo_url IS NULL")
     fun getPlanterIdentificationsToUpload(): List<PlanterIdentificationsEntity>
 
+    @Transaction
+    @Query("SELECT * FROM planter_identifications WHERE identifier = :identifier")
+    fun getPlanterIDByIdentifier(identifier: String?): Int
```
mozarty: I don't know if you resolved this issue already, but this query `SELECT *` will return All fields so you can change the return type to `PlanterIdentificationsEntity` OR change the query to `select planterDetailsId` and it will match the return type

deepwinter:
@LaviniaDragunoi  I am unable to proceed past the identification screen when I run this branch.

LaviniaDragunoi:
Yes, @deepwinter  is not ready. Still working on this. I have the full support of the team for this and I hope to have real progress soon.

app/src/main/java/org/greenstand/android/TreeTracker/database/dao/PlanterDao.kt
LaviniaDragunoi: Thanks, @mozarty , I am updating now the code with the latest changes to use the TreeManager instead this way I've done here. 
I'll come back with the update and then if you have some time to review my code it will be great!

app/src/main/java/org/greenstand/android/TreeTracker/database/dao/PlanterDao.kt
```diff
@@ -21,6 +22,13 @@ interface PlanterDao {
     @Query("SELECT * FROM planter_identifications WHERE photo_url IS NULL")
     fun getPlanterIdentificationsToUpload(): List<PlanterIdentificationsEntity>
 
+    @Transaction
+    @Query("SELECT * FROM planter_details WHERE identifier = :identifier")
+    fun getPlanterDetailsByIdentifier(identifier: String): LiveData<PlanterDetailsEntity>
```
Elforama: Using LiveData usually means that we are getting updates of some type of data, in this case PlanterDetailsEntity. I'm guessing we only need to load this once with no need for updates. If that's the case, we can just return a plain PlanterDetailsEntity

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,182 @@
...
+import org.greenstand.android.TreeTracker.utilities.ValueHelper.PHONE_NUMBER
+import org.greenstand.android.TreeTracker.viewmodels.PlanterDetailsViewModel
+import timber.log.Timber
+
+class LoginFragment : Fragment(){
+    var phoneNumberEntered: String? = null
+    var emailEntered: String? = null
+    private var mPhotoPath: String? = null
+    lateinit var viewModel: PlanterDetailsViewModel
+    private var fragment: Fragment? = null
```
Elforama: I don't think fragment and fragmentTransaction properties are needed

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,182 @@
...
+            val fragmentTransaction = activity?.supportFragmentManager?.beginTransaction()
+            fragmentTransaction?.addToBackStack(null)?.replace(R.id.containerFragment, termsFragment)
+            fragmentTransaction?.commit()
+
+        }
+    }
+@SuppressLint("NewApi")
+fun activateLoginButton(){
+    login_button.apply {
+        setTextAppearance(R.style.ActiveButtonStyle)
```
Elforama: There is a way to set enabled/disables styles on a view in xml, then you'd only have to say login_button.enabled = true/false

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,182 @@
...
+            override fun beforeTextChanged(p0: CharSequence?, p1: Int, p2: Int, p3: Int) {
+                inactivateLoginButton()
+            }
+            override fun onTextChanged(p0: CharSequence?, p1: Int, p2: Int, p3: Int) {
+                inactivateLoginButton()
+            }
+
+            override fun afterTextChanged(editTextInputs: Editable) {
+                phoneNumberEntered = editTextInputs.toString()
+                viewModel.isUserPresentOnDevice(phoneNumberEntered!!).observe(this@LoginFragment,
```
Elforama: You should not do any observing inside any of these callbacks for the edit text. This is creating a new observer every time the text changes. Instead we can make a function in the viewmodel called something like 
`fun updatePhoneNumber(phoneNumber: String)`
That can be called each time text is updated. Then, outside of any callbacks, you can observer if the user is present on the device. The method `isUserPresentOnDevice` doesn't have to take any input, since the `updatePhoneNumber` will tell the viewmodel what was input by the user.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/LoginFragment.kt
```diff
@@ -0,0 +1,182 @@
...
+                    Observer {
+                        if (it != null && it.identifier == phoneNumberEntered) {
+                            activateLoginButton()
+                        } else {
+                            sign_up_button.visibility = View.VISIBLE
+                            emailEditText.addTextChangedListener(object : TextWatcher {
```
Elforama: You can move this outside of all these callbacks, and listen to textchange events no matter if it's showing or not. You can update the email address and observe if the user is on device just like above.

app/src/main/java/org/greenstand/android/TreeTracker/database/dao/PlanterDao.kt
LaviniaDragunoi: I tried this approach but I have a bug and I had struggled with it for two days. I can't figure out what I'm doing wrong. You can see in my last commit what I've done. When you have some time maybe you could take a look. Thank you so much!

Elforama: If it doesn't crash the app we can probably merge what you have to allow additional people to figure out the bugs. Sound good @LaviniaDragunoi @deepwinter ?

LaviniaDragunoi: The last commit is crashing the app. I could change it to the old version of my  code that is not the best approach but is not crashing the app. So I think we should wait a little with the merge.

Elforama: Changing it to the old version and getting this merged sounds like a good idea.

LaviniaDragunoi:
@Elforama now the app is not crashing.
Problems: 
- The app is not writing in the planter_identifications DB's table
- The menu bar icon has to disappear from the fragments  that are for the login/signup part
- The user's details/identification is not sent to the MapFragment
- Permissions are not customized as the UX Desing requires, from official documentation I understood that this change is not possible( If you have some idea how this could be done.. it will be great - I tried to a certain point with a customized Dialog popup and so on..).
If you could take a look I would appreciate
Thanks.

PR #244, opened 2019-03-26
Fixed to send values for tree note, fixed to use appropriate coroutin…

deepwinter:
…e context to update UI

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/database/dao/TreeDto.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/database/dao/TreeDto.kt
@@ -12,7 +12,6 @@ class TreeDto {
     var name: String? = null
     var note: String? = null
     var isOutdated: Boolean? = false
-    var content: String? = null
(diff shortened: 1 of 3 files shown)


deepwinter:
Ignore the bitrise failure above, it is google-services.json related.  Fix elsewhere.

app/src/main/java/org/greenstand/android/TreeTracker/database/dao/TreeDto.kt
```diff
@@ -10,6 +10,7 @@ class TreeDto {
     var longitude: Double = 0.0
     var accuracy: Int = 0
     var name: String? = null
+    var note: String? = null
```
Elforama: Wasn't the note previously the content value on line 15? If that's the case, it should be removed.

deepwinter: Yup, I changed the name elsewhere but missed it in the model I guess.

Issue #247, opened 2019-03-27
Remove "Update" text on Data screen


Davidezrajay:
This field is currently not supported and should be hidden or removed.

PR #249, opened 2019-03-29
247 remove update text

Elforama:
Issue: https://github.com/Greenstand/treetracker-android/issues/247


deepwinter:
Ya totally, update will come eventually, but having it in the codebase is just confusing since we have nothing in place to support this feature.

PR #253, opened 2019-04-05
127 tree UUID

Elforama:
Time Spent: 2 hrs

Notes: 

- Adds a UUID to every new tree db entry.
- Sends UUID to server with rest of tree payload
- Updated tree attributes table to be a key/value table to allow generic attributes to be added to a tree


deepwinter:
Looks good to me.

PR #260, opened 2019-04-24
New login design

Elforama:
@LaviniaDragunoi and @deepwinter Please test this thoroughly. 

If user already has their email/phone number saved then the flow is login -> camera -> map
If user doesn't have their email/phone saved, they go to login -> signup -> terms -> camera -> map
If a user selects "change user" from the menu option, they are brought to the login screen


deepwinter:
Awesome PR !

PR #261, opened 2019-04-25
Give camera preview a title bar and make email not required

deepwinter:


app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.kt
```diff
@@ -60,6 +62,9 @@ class CameraActivity : Activity(), Camera.PictureCallback, View.OnClickListener,
         intent.extras?.let {
             captureSelfie = intent.extras!!.getBoolean(ValueHelper.TAKE_SELFIE_EXTRA, false)
         }
+
+        setSupportActionBar(camera_toolbar)
+        supportActionBar?.title = "Take a Selfie"
```
Elforama: Can we move this to a string resource? Better yet, just add this as a string resource to the xml.

deepwinter: Right, I actually did that in the XML.  Thanks.

app/src/main/res/layout/activity_camera.xml
```diff
@@ -2,10 +2,37 @@
...
+            android:minHeight="?attr/actionBarSize"
+            style="@style/TextTitleStyle">
+
+            <TextView
+                android:id="@+id/toolbarTitle"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:layout_gravity="center"
+                android:textSize="@dimen/button_text_size"
+                tools:text="Take a Selfie" />
```
Elforama: This needs to be android:text or else it only shows up in the editor preview.

PR #262, opened 2019-04-25
Send attributes as key value

deepwinter:


app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/DataViewModel.kt
```diff
@@ -278,13 +278,10 @@ class DataViewModel : CoroutineViewModel() {
...
         val imageUrl = getImageUrl() ?: return null
 
-        val attributesRequest = TreeManager.getTreeAttribute(treeDto.tree_id, TreeManager.TREE_COLOR_ATTR_KEY)?.let {
-            AttributesRequest(
-                heightColor = it,
-                appVersion = TreeManager.getTreeAttribute(treeDto.tree_id, TreeManager.APP_VERSION_ATTR_KEY)!!,
-                appBuild = TreeManager.getTreeAttribute(treeDto.tree_id, TreeManager.APP_BUILD_ATTR_KEY)!!,
-                flavorId = TreeManager.getTreeAttribute(treeDto.tree_id, TreeManager.APP_FLAVOR_ATTR_KEY)!!
-            )
+        val attributesList = TreeTrackerApplication.getAppDatabase().treeAttributesDao().getTreeAttributesByTree(treeDto.tree_id)
```
Elforama: Can we add this to TreeManager

deepwinter: Yeah, I wasn't immediately sure how to do it.  Actually other 'get' queries are also not included in tree manager, so I wasn't sure if there was a reason to use one pattern over the other.

Elforama: You can look at TreeManager.getTreeAttribute() as an example. 
Not all DB operations for trees are in the TreeManager yet, it's still a WIP to consolidate everything.

deepwinter: Great, thanks.

deepwinter: Pushed the fix.

PR #265, opened 2019-04-26
264 extend input fields

Elforama:
Fix login/signup input fields


deepwinter:
Ah, so the input fields themselves were not a fixed width.  Should we always be using the material UI types instead of EditText, etc from this point forward?

PR #272, opened 2019-04-30
Change the Add Tree Button

Davidezrajay:
Replaced <solid android:color="@color/greenColor" />
With Just Dig It's color code <solid android:color="#7AC144" />


deepwinter:
I can't allow you to do it this way.  You need to create a colors entry in colors.xml for each project for this button, and then use it here

deepwinter:
The way you have done this changes it in both versions of the app.

deepwinter:
Rejecting this PR.

Issue #246, opened 2019-03-27
GPS ACCURACY not displayed


sebastiangaertner:
Accuracy is not showing on the initial map (under ADD TREE button). 
The app is recording accuracy but not displaying it.

Davidezrajay:
https://github.com/Greenstand/treetracker-android/issues/172

deepwinter:
from Ezra:
It doesn’t seem to turn green.. stays red but allows you to take the photo... photo shows up with a god accuracy of 6 or so

deepwinter:
@Elforama I see this happen after I add a tree.  Before adding a tree I have accuracy data, afterwards its not reappearing.

deepwinter:
In MainActivity we have

`        if (fragmentMapGpsAccuracy != null) {
`

Since we are updating the accuracy text on the maps fragment from the main activity, and we know we are still getting location updates, I am guessing that something along the lines of this variable going null incorrectly is resulting in the accuracy not updating on the maps view.  Or at least this is the part of the code to look at.

This bug pops up after you have added a few trees.

deepwinter:
This needs to be fix, since the accuracy gets stuck on red even when it should be green.

deepwinter:
well, the view variables for GPS info are not null, but though they are being set in code, they are not updating.  It's not clear to me why this would be.

PR #279, opened 2019-05-01
246 accuracy fix

Elforama:
Resolves: https://github.com/Greenstand/treetracker-android/issues/246


app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
```diff
@@ -167,15 +167,10 @@ class MapsFragment : androidx.fragment.app.Fragment(), OnClickListener, OnMarker
...
 
-        var v: View? = view
-
-        try {
-            v = inflater.inflate(R.layout.fragment_map, container, false)
-        } catch (e: InflateException) {
-            Timber.d(e.localizedMessage);
-        }
-
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
```
mckornfield: So the fix was to use a different lifecycle method for the rest of this?

app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
```diff
@@ -195,10 +190,9 @@ class MapsFragment : androidx.fragment.app.Fragment(), OnClickListener, OnMarker
 
         (activity as AppCompatActivity).supportActionBar!!.setDisplayHomeAsUpEnabled(false)
 
-        val fab = v!!.addTreeButton
-        fab.setOnClickListener(this)
-        if (BuildConfig.BUILD_TYPE === "dev") {
-            fab.setOnLongClickListener(this)
+        addTreeButton.setOnClickListener(this)
+        if (FeatureFlags.DEBUG_ENABLED) {
```
mckornfield: Good cleanup

app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
Elforama: Seems like that's what was needed.

deepwinter:
Looks fine to me!

PR #289, opened 2019-05-14
Add dependency injection

Elforama:
Notes:
Adds dependency on Koin DI library
Updated classes to use constructor injection to allow easy testing in the future
Created modules that provide dependencies


deepwinter:
Looks great!

PR #294, opened 2019-05-16
284 tag user login with location

Elforama:
Issue: https://github.com/Greenstand/treetracker-android/issues/284

This PR saves location data (lat,long) with user registation and sends it when uploading data.
Location data is send with the key "location" and the value "lat,long" e.g. "12.22,15.52"

Due to changes in the migration, please run an older version of the app before the tree height feature was added, then after adding a tree, update the app to this current version to ensure there are no migration bugs.


deepwinter:
Hey Jonathon, is there a reason why we are changing the migration instead of just adding a new migration on top of the existing ones?   Could we follow that pattern?

Elforama:
We're still on v2, the app has not been shipped since changes were made. Also, less work for me.

deepwinter:
It's been pushed out on Fabric though, will this cause issues for Ezra or anyone who has loaded it for QC already, since they won't be able to backtrack so easily.  We can probably just tell them to delete and reinstall, but I am know that Ezra likes to share around our QC build.

deepwinter:
Great work otherwise!  We'll just wait for @Davidezrajay to weigh with who is accessing the QC builds and then I can merge this.

PR #296, opened 2019-05-16
Delete nature.jpg

Davidezrajay:


deepwinter:
so much for nature.jpg

PR #303, opened 2019-05-19
Fix map battery

deepwinter:


deepwinter:
@Elforama looking for a fix for the battery issue referenced in #android.  Wondering if you have thoughts here.

app/build.gradle
```diff
@@ -106,7 +106,7 @@ android {
             applicationIdSuffix ".prerelease"
             resValue "string", "app_name", "Tree Tracker"
             buildConfigField "Boolean", "GPS_ACCURACY", "true"
-            buildConfigField 'String', 'BASE_URL', '"https://treetracker.org/api/mobile/"'
+            buildConfigField 'String', 'BASE_URL', '"http://67.205.142.69/api/mobile/"'
```
Elforama: Why was this changed?

app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
```diff
@@ -93,6 +93,10 @@ class MapsFragment : androidx.fragment.app.Fragment(), OnClickListener, OnMarker
     override fun onPause() {
         super.onPause()
 
+        if(this.map != null){
+            this.map?.isMyLocationEnabled = false
```
Elforama: Is onMapReady get called again when the app resumes? If not, the location updates might not trigger again.

deepwinter: Yes, getMapAsync in onResume takes care of that (onMapReady is the async callback)

app/build.gradle
deepwinter: Oh snap, thanks, that was something I while testing out the SSL on treetracker.org.  build.gradle gets auto-committed with its version bump when I push a new build, so this slipped in.

PR #313, opened 2019-05-28
Fix lag

Elforama:
This fixes the lag that appears when going to another screen and then back to the map.


deepwinter:
Cool, so it looks like that fragment is leaking a context, and as long as we don't destroy it we are OK.

But the leaked context will still be a problem if the activity gets recreated.. Agree/disagree ?

Elforama:
The fragment is bound to this activity. So it shouldn't be able to leak of its container is destroyed

PR #322, opened 2019-06-04
Add focus metric

DominionSoftware:
I added the focus metric code. It looks at the taken image and adds a red frame and text around the image if the focus metric is less than 950.0. I wasn't able to get my tablet to be out of focus very easily, so I added Unit test code (commented out) for testing.

I added strings to the resources and used google translate for swahili.

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.kt
@@ -307,6 +307,22 @@ class CameraActivity : AppCompatActivity(), Camera.PictureCallback, View.OnClick
         return 0.0;
     }
 
+    /**
+     * TODO make a gaussian kernel
+     */
+    private fun makeKernel(rows:Int, cols: Int) :  Array<Array<Int>>{
+        var result = Array(rows) { Array(cols) { 0 } }
+        return result
+    }
+    /**
+     *  TODO To test the focus detection, we need to blur some images.
+     */
+    private fun applyKernel(image: Array<Array<Int>>,rows: Int,cols: Int,kernek: Array<Array<Int>>,kRows: Int,kCols: Int ) {
+
+
+    }
+
+
     private fun setPic() {
 
         /* There isn't enough memory to open up more than a couple camera photos */
--- a/app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.kt
@@ -51,7 +51,6 @@ class CameraActivity : AppCompatActivity(), Camera.PictureCallback, View.OnClick
     private var captureSelfie: Boolean = false
 
 
-
     public override fun onCreate(savedInstanceState: Bundle?) {
         super.onCreate(savedInstanceState)
         setContentView(R.layout.activity_camera)
@@ -70,7 +69,7 @@ class CameraActivity : AppCompatActivity(), Camera.PictureCallback, View.OnClick
 
         setSupportActionBar(camera_toolbar)
         supportActionBar?.title = ""
-        if(captureSelfie) {
+        if (captureSelfie) {
             cameraToolbarTitle.text = getString(R.string.take_a_selfie)
         } else {
             cameraToolbarTitle.text = getString(R.string.add_a_tree)
@@ -87,10 +86,10 @@ class CameraActivity : AppCompatActivity(), Camera.PictureCallback, View.OnClick
 
             captureButton?.visibility = View.INVISIBLE
 
-            while(mCamera == null){
+            while (mCamera == null) {
                 try {
 
-                    if(captureSelfie) {
+                    if (captureSelfie) {
 
                         val numberOfCameras = Camera.getNumberOfCameras()
                         if (numberOfCameras > 1) {
@@ -116,6 +115,7 @@ class CameraActivity : AppCompatActivity(), Camera.PictureCallback, View.OnClick
             captureButton?.visibility = View.VISIBLE
         }
     }
+
     /** Check if this device has a camera  */
     private fun checkCameraHardware(context: Context): Boolean {
         return if (context.packageManager.hasSystemFeature(PackageManager.FEATURE_CAMERA)) {
@@ -146,7 +146,7 @@ class CameraActivity : AppCompatActivity(), Camera.PictureCallback, View.OnClick
             e.printStackTrace()
         }
 
-        if(captureSelfie){
+        if (captureSelfie) {
 
             var exif: ExifInterface? = null
             try {
@@ -169,7 +169,7 @@ class CameraActivity : AppCompatActivity(), Camera.PictureCallback, View.OnClick
 
     private fun compressImage() {
 
-        if(mCurrentPhotoPath != null) {
+        if (mCurrentPhotoPath != null) {
             val photo = Utils.resizedImage(mCurrentPhotoPath!!)
 
             val bytes = ByteArrayOutputStream()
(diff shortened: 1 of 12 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.kt
```diff
@@ -198,10 +206,109 @@ class CameraActivity : AppCompatActivity(), Camera.PictureCallback, View.OnClick
...
 
+    /**
+     * // Compute average of sum of squares of the gradient in H and V directions.
+     */
+    private fun brennersFocusMetric(input: Array<Array<Int>>): Double {
+
+        val rows = input.lastIndex + 1;
+        val cols = input.get(0).lastIndex + 1;
+
+        var V = Array(rows) { Array(cols) { 0 } }
```
Elforama: Can we please use lower case for variable names?

app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.kt
```diff
@@ -198,10 +206,109 @@ class CameraActivity : AppCompatActivity(), Camera.PictureCallback, View.OnClick
...
+                val grad = input[row + 2][col] - input[row][col];
+                V[row][col] = grad;
+            }
+        }
+
+        var sum = 0;
+        for (row in 0 until rows) {
+            for (col in 0 until cols) {
+                val HRC = H[row][col];
+                val VRC = V[row][col];
```
Elforama: Same as above

app/src/main/java/org/greenstand/android/TreeTracker/fragments/NewTreeFragment.kt
```diff
@@ -225,6 +233,17 @@ class NewTreeFragment : androidx.fragment.app.Fragment(), OnClickListener, Activ
                     }
 
                     setPic()
+                    if (MainActivity.mImageQuality < FOCUS_THRESHOLD) {
```
Elforama: Instead of adding a new property into the MainActivty, can we instead pass back the image quality value using onActivityResult?

app/src/main/java/org/greenstand/android/TreeTracker/fragments/NewTreeFragment.kt
```diff
@@ -85,6 +88,11 @@ class NewTreeFragment : androidx.fragment.app.Fragment(), OnClickListener, Activ
 
         mImageView = v.fragmentNewTreeImage
 
+        frameLayout = v.fragment_new_tree_frame_layout
+
+        focusWarning = v.fragment_new_tree_focus_warning_text
```
Elforama: These local references to the view are not needed. We can directly use fragment_new_tree_frame_layout and fragment_new_tree_focus_warning_text without using `v`

DominionSoftware:
mImageView already existed so I took that as a model.

How does one do this ?

Elforama:
Any place where you use focusWarning, replace it with fragment_new_tree_focus_warning_text. And any place you use frameLayout, replace it with fragment_new_tree_frame_layout. Those long underscore names are part of a plugin called Kotlin Synthetic Views. It allows us to directly reference views from a xml layout. You can read more about it here if you're interested: https://kotlinlang.org/docs/tutorials/android-plugin.html

Feel free to update the imageView to do the same.

app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.kt
```diff
@@ -328,6 +434,98 @@ class CameraActivity : AppCompatActivity(), Camera.PictureCallback, View.OnClick
     }
 
 
+/** UNIT TEST CODE FOR FOCUS METRIC
+    private fun printImage(image: Array<Array<Int>> ) {
```
Elforama: What do you think of moving this testing code into its own file?

DominionSoftware:
That's fine. Where should the file live? Are there other unit test like files?




Rick Frank
Dominion Software, Inc.
825 Beacon Street
Newton, MA 02459
Medical, Scientific, and Industrial Software

On Jun 4, 2019, at 9:33 PM, Jonathan Muller <notifications@github.com<mailto:notifications@github.com>> wrote:


@Elforama commented on this pull request.

________________________________

In app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.kt<https://github.com/Greenstand/treetracker-android/pull/322#discussion_r290554752>:

> @@ -328,6 +434,98 @@ class CameraActivity : AppCompatActivity(), Camera.PictureCallback, View.OnClick
     }


+/** UNIT TEST CODE FOR FOCUS METRIC
+    private fun printImage(image: Array<Array<Int>> ) {


What do you think of moving this testing code into its own file?

DominionSoftware:
OK I moved all the image processing code to ImageUtils.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/NewTreeFragment.kt
DominionSoftware: Ok this is done.

app/src/main/java/org/greenstand/android/TreeTracker/activities/CameraActivity.kt
DominionSoftware: I move everything to ImageUtils

Elforama:
@DominionSoftware Looks like there are some minor conflicts with the strings file.

DominionSoftware:
Ok string conflict resolved

Elforama:
@DominionSoftware I've tried running this on my Pixel 3, but once I take a picture the app freezes for a while before crashing. Any idea what's going on?

DominionSoftware:
No idea without more information. I guess I need to get the phone from Ezra. ( he's sending me one ) I only have a samsung tablet to test with.


Rick Frank
Dominion Software, Inc.
825 Beacon Street
Newton, MA 02459
Medical, Scientific, and Industrial Software

On Jun 9, 2019, at 12:00 PM, Jonathan Muller <notifications@github.com<mailto:notifications@github.com>> wrote:


@DominionSoftware<https://github.com/DominionSoftware> I've tried running this on my Pixel 3, but once I take a picture the app freezes for a while before crashing. Any idea what's going on?

Elforama:
I just got it working with my emulator, however it's always showing the image as blurry even when I've ensured the image was sharp. Are you seeing this behavior at all?

build.gradle
```diff
@@ -7,7 +7,7 @@ buildscript {
         jcenter()
     }
     dependencies {
-        classpath 'com.android.tools.build:gradle:3.3.1'
+        classpath 'com.android.tools.build:gradle:3.4.1'
```
Elforama: This has caused some build issues, can we change it back?

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Wed Nov 28 08:54:44 EET 2018
+#Wed May 29 15:03:23 EDT 2019
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.1-all.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-5.1.1-all.zip
```
Elforama: Same as above

Elforama:
Here is the user flow for blur detection: https://docs.google.com/drawings/d/1CkkffY48rDrCJ3hLVtdVa4diHrFUj5D3oAzlLWuU_Ec/edit

There doesn't yet seem to be a way to retake the photo.

@peacengn How should we allow the user to retake the photo? By tapping on the photo?

build.gradle
DominionSoftware: I don't have any reason to not change it back

DominionSoftware: The way I tested it was to blur images using the gaussian filter I put into ImageUtiles and then measure the metric compared to non blurred images, the result being empirically tested to be metric of 950.0

DominionSoftware: image
Current UI without trash can

DominionSoftware:
> I just got it working with my emulator, however it's always showing the image as blurry even when I've ensured the image was sharp. Are you seeing this behavior at all?

No.

DominionSoftware:
I limited the image size for the focus test to 200 pixels wide. I changed the color to be Yellow, as a caution, rather than red. Because if you take a picture of an in focus blank wall, it will say it's out of focus (because it's low contrast). So it should be a warning / reminder and not a fail. I changed the string to be:

"We've detected that your photo may be blurry. Please push the back button retake the photo, if necessary.

I also made the focus threshold slightly lower (permissive) based on some tests. I did some tests with my tablet and Pixel 3A XL cameras and it seemed to be working as expected. Some tests with less capable phones are needed.

DominionSoftware:
I framed the text "Caution Yellow" instead of the image as I could not get the image to frame tightly.

The blur detection code adds no noticeable performance hit to my phone or tablet. However, they are fairly new an have sufficient RAM.


Blurred Image test (used a piece of cellophane to blur the image); non blurred image follows after.


Screenshot_20190611-170500



Screenshot_20190611-170359

DominionSoftware:
Closing pull request per conversations with Ezra and Sebastian. Code can stay in branch if future use becomes desired.

PR #340, opened 2019-06-16
312 gps date

Elforama:
This is a cleanup for location updates and tree creation. 
Phones running android 8 or higher can sync data.
A month of time is no longer added to time stamps


deepwinter:
This looks great.  I'm so glad you created the location manager.

PR #381, opened 2019-08-18
Send an app instance ID in updateDevice API call

hz97:
Add instance id to DeviceRequest


app/src/main/java/org/greenstand/android/TreeTracker/api/models/requests/DeviceRequest.kt
```diff
@@ -23,4 +24,7 @@ data class DeviceRequest(@SerializedName("app_version")
                          @SerializedName("androidRelease")
                          val androidRelease: String = Build.VERSION.RELEASE,
                          @SerializedName("androidSdkVersion")
-                         val androidSdkVersion: Int = Build.VERSION.SDK_INT)
\ No newline at end of file
+                         val androidSdkVersion: Int = Build.VERSION.SDK_INT,
+                         @SerializedName("instance_id")
+                         val instance_id: String = FirebaseInstanceId.getInstance().id
```
Elforama: Great job! Can we just make sure this field is camelCase like the rest?

hz97: Yes! Just changed it!

PR #382, opened 2019-08-18
Add Swahili Privacy Policy - GreenStand flavor

hz97:


Elforama:
@hz97 This is good, but now that I think of it, this would be best to be in the main (non flavor) strings file instead of the Greenstand flavor. That will allow all other flavors to get this translation when the phone is set to it.

hz97:
> @hz97 This is good, but now that I think of it, this would be best to be in the main (non flavor) strings file instead of the Greenstand flavor. That will allow all other flavors to get this translation when the phone is set to it.

Done!

Elforama:
Thanks!

PR #387, opened 2019-08-24
Remove menu from login flow

Elforama:
The menu was showing in the toolbar when changing users, this has lead to some recent crashes. The menu only shows on the map screen now.


deepwinter:
Good to go.

PR #393, opened 2019-09-09
Make tree uploading implement strategy pattern

darrengyl:


Elforama:
Thanks Darren, great idea!

app/src/main/java/org/greenstand/android/TreeTracker/di/AppModule.kt
```diff
@@ -85,11 +86,21 @@ val appModule = module {
...
-    factory { SyncDataUseCase(get(), get(), get(), get()) }
+    factory(named(SyncDataUseCase.BUNDLE_UPLOAD)) { SyncDataUseCase(
+        treeLoadStrategy = BundleTreeUploadStrategy(uploadTreeBundleUseCase = get()),
+        uploadPlanterDetailsUseCase = get(),
+        api = get(),
+        dao = get()
+    ) }
+
+    factory(named(SyncDataUseCase.CONTINUOUS_UPLOAD)) { SyncDataUseCase(
+        treeLoadStrategy = ContinuousTreeUploadStrategy(syncTreeUseCase = get()),
```
Elforama: Can we create factory methods for the two upload strategies and just use get() when creating the SyncDataUsecases?

darrengyl: Done

PR #415, opened 2020-02-15
Fix app crashes when adding many trees in quick succession

arunbakt:
The change fixes the issue where the app crashes when trying to add new
trees in quick succession. The crash is caused by the IllegalArgument exception.
See the exception pasted below. The fix is to disable the add new tree
button as soon after the very first click is registered.

Due to potential sluggishness of writing the new tree data to db and rendering the map
fragment screen, a quick succession of clicks on Add tree button by the user triggers its onClick
listeners more than once which cause the navigation action mapsFragment to new tree graph to take place.
If the navigation action has already completed, the duplicate action to navigate to the same destination
throws the exception causing the app to crash.

Steps to reproduce the crash:
1. Add new tree from the app.
2. The camera activity is launched.
3. Take an image which starts the new tree fragment.
4. Click save button which saves the data to the db and takes you to map
fragment
5. Immediately click on add tree button several times.
6. After few repititions of the above steps, the app crashes.

Verification for the fix:
1. Repeat the same steps listed above to crash the app.
2. The add tree button is disabled after the very first click.
3. No crash occurs.

```

    --------- beginning of crash
E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.greenstand.android.TreeTracker.dev, PID: 22273
    java.lang.IllegalArgumentException: navigation destination
org.greenstand.android.TreeTracker.dev:id/action_mapsFragment_to_new_tree_graph
is unknown to this NavController
        at
androidx.navigation.NavController.navigate(NavController.java:789)
        at
androidx.navigation.NavController.navigate(NavController.java:730)
        at
androidx.navigation.NavController.navigate(NavController.java:716)
        at
androidx.navigation.NavController.navigate(NavController.java:907)
        at
org.greenstand.android.TreeTracker.fragments.MapsFragment.onClick(MapsFragment.kt:155)
        at android.view.View.performClick(View.java:7346)
        at android.widget.TextView.performClick(TextView.java:14177)
 ```

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
@@ -141,15 +141,14 @@ class MapsFragment : androidx.fragment.app.Fragment(), OnClickListener, OnMapRea
 
     override fun onClick(v: View) {
         v.vibrate()
-        // Disable the addTreeButton below to avoid triggering the onClick listener
-        // more than one once
-        (v.findViewById(R.id.addTreeButton) as Button)?.isEnabled = false
         when (v.id) {
             R.id.addTreeButton -> {
                 Timber.d("fab click")
 
                 if (userLocationManager.hasSufficientAccuracy() || !FeatureFlags.HIGH_GPS_ACCURACY) {
-
+                    // Disable the addTreeButton below to avoid triggering the onClick listener
+                    // more than one once.
+                    addTreeButton.isEnabled = false
                     val currentTimestamp = System.currentTimeMillis() / 1000
                     val lastTimeStamp = sharedPreferences.getLong(ValueHelper.TIME_OF_LAST_PLANTER_CHECK_IN_SECONDS, 0)
                     if (currentTimestamp - lastTimeStamp > ValueHelper.CHECK_IN_TIMEOUT) {
--- a/app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
@@ -12,6 +12,7 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.View.OnClickListener
 import android.view.ViewGroup
+import android.widget.Button
 import android.widget.TextView
 import android.widget.Toast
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt [hunk 1]
```diff
@@ -140,7 +141,9 @@ class MapsFragment : androidx.fragment.app.Fragment(), OnClickListener, OnMapRea
 
     override fun onClick(v: View) {
         v.vibrate()
-
+        // Disable the addTreeButton below to avoid triggering the onClick listener
+        // more than one once
+        (v.findViewById(R.id.addTreeButton) as Button)?.isEnabled = false
```
Elforama: This disables the button right after it's clicked. What happens if we want to add another tree? Will the button still be disabled?
If that's the case, what about adding a delay between clicks: https://stackoverflow.com/questions/16534369/avoid-button-multiple-rapid-clicks?rq=1

arunbakt:
The button is enabled when the tree is saved and the MapFragment is
navigated to. So, the user is able to add the tree again. I tested that.

I believe even if we introduce a delay, if the app was still processing the
frames, it would cause duplicate onClick process to execute the navigate
action
while on the NewFragment causing similar exception.

Arun

app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
arunbakt: The button is enabled when the tree is saved and the MapFragment is navigated to. So, the user is able to add the tree again. I tested that.

I believe even if we introduce a delay, if the app was still processing the frames, it would cause duplicate onClick process to execute the navigate action
while on the NewFragment causing similar exception.

Elforama: Thanks, this looks good then. I'd just like to ask if you can reuse the view reference we already have for that button instead of finding it again, and moving it to the check for when only the addTreeButton is clicked.

arunbakt: Thank you for the feedback. I updated the code (and tested the issue is resolved) as per your suggestions.

PR #419, opened 2020-02-17
Add action to travis file to succesfully run PR builds

jhoon:
The tar file is available on all branches that are up to date,
even master is failing if we didn't have the step where we
extract the info from the secrets.tar file.


arunbakt:
Looks good to me and we could use this until we further revise this to enable tests and builds runs soon as part of PR builds.

PR #420, opened 2020-02-17
This is a test PR

jhoon:
No need to merge!

This is just to test some occurrences on Travis


jhoon:
So, from what I've ended up reading (https://docs.travis-ci.com/user/pull-requests/) seems like they recommend just skipping all PR builds from external contributors if the project depends on env variables for encryption.

One strategy could be to send PRs to a throwaway branch in this repo and then open a PR from there, since those have access to said variables. Might be a great idea for some specific tasks, so we are not always asking people to collaborate differently.

deepwinter:
To start we could just have a fake set of credentials that we inject to PR builds so that we can at least have build ability of PRs automatically tested.

I think what we could do longer term is have a fast lane script that pulls some credentials into the build dynamically when it runs.  These could be linked to a part of our cloud that we don’t need to be especially concerned about securing.  It would be really useful to have external PRs automatically build for testing, without having to do any special work moving PRs/branches around.

Does this seem to complicated though, or doable?

jhoon:
Yeah, I actually created mock credential files and added those to a mock `secrets.tar` file. This will be used for external PR builds and replaced anytime the build comes from the repository itself.

This was done by making the decryption step skippable when it comes to external PRs.

PR #426, opened 2020-03-04
Update strings.xml

sebastiangaertner:
Please review. These are small changes mainly sorting out the spelling of Treetracker which should be one word across. The only line i swas not sure about is line 4.  this looks like a comment to me so please confirm. I will work on the formatting of the Legal Screen and the About screen as well. These are cosmetic but important changes.

### DIFF
--- a/app/src/main/res/values/strings.xml
+++ b/app/src/main/res/values/strings.xml
@@ -119,18 +119,18 @@
     <string name="location_permission_alert_title">Allow Treetracker to access location?</string>
     <string name="location_permission_alert_message">Please open Settings and turn on location permission for Treetracker</string>
 
-    <string name="information_text">
-        •	This tree tracking planter app is made to track planted trees while supporting people and organizations that plant, monitor and protect trees.
-        
-        •	Your app requires 10 meters of accuracy to take a photo.
-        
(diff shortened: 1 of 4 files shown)


sebastiangaertner:
We should also format the policy_text_blob and the information_text both rendered as WebView

since this https://developer.android.com/guide/topics/resources/string-resource.html#FormattingAndStyling is very limiting.

Tutorial: https://alvinalexander.com/android/how-show-html-string-in-android-textview-webview

sebastiangaertner:
I have also changed the Gradle file and in the app name changed the name from Tree Tracker to Treetracker. This should only effect the display of it not any package names.

app/src/main/res/values/strings.xml
```diff
@@ -208,85 +208,35 @@
...
-
-It is our intent to hold the principles of this policy, but we do understand there may be a need to change it from time to time. Changes will be posted on our website as soon as possible.
-        \n
-        \n
-
-© 2019 GreenStand (registered 501(c)3 nonprofit. EIN: 47-3150283 ). Info@greenstand.org
-        \n
-Privacy Policy
-
+	    <![CDATA[
```
Elforama: Please undo this HTML change. This should be done in a different PR. This requires code changes.

Issue #416, opened 2020-02-16
UI Update for Tree Data screen


Elforama:
This updated screen will contain new data implemented from endpoint in: https://github.com/Greenstand/treetracker-android/issues/417

The sync/stop button will be replaced with a new button that contains a cloud icon and X icon for the two states.

This screen should function when offline and will require updates to the database to store payment info from the new endpoint.

<img width="688" alt="Screen Shot 2020-02-16 at 6 30 13 PM" src="https://user-images.githubusercontent.com/6439445/74614983-8e687080-50ea-11ea-953c-b1b394474f41.png">

arunbakt:
I have a question regarding the upload button where when a user clicks on it, it starts the sync process and changes the button to show a cross according to the requirement.  

1. Should we give the ability to cancel the upload process via the button with the 'X' icon? Assuming the upload of the local tree data and download of payment data and validated trees happening in two distinct phases,  we could trigger the cancellation safely. Wanted to confirm the user experience.

2. Trying to use a 'X' icon from the material design site, looks weird for my eyes. I suggest we use a solid stop icon instead. Let me know what you think.

Davidezrajay:
A black Square is good. Thanks for your idea.

PR #428, opened 2020-03-12
Tree Data UI changes

arunbakt:
The following PR contains changes addressing the requirement specified
in
https://github.com/Greenstand/treetracker-android/issues/416

- Upload, validated tree counts and payment data are modified to show
information from greenstand servers instead of local db values
- Waiting to upload tree count is obtained from the local db
- A new table to store the server  returned account data (upload,
validated, payment data) is created with planterInfoId as its primary
key

Note: Currently Uploaded, validated counts and payment data returns zero
as the value since the integration with the API is not yet done.

Tests:

Uploading to cloud
- Login to the app and add new trees
- Go to tree data and on clicking upload to cloud button, the number of
trees to upload goes down to zero.

Canceling the upload process
- After capture and add new trees starting the upload process
- While the process is on going, click on the stop icon
- The tree count of pending to upload will not change

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/DataViewModel.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/DataViewModel.kt
@@ -13,7 +13,6 @@ import org.greenstand.android.TreeTracker.managers.UserManager
 import java.util.concurrent.TimeUnit
 import kotlin.properties.Delegates
 
-
 class DataViewModel(private val userManager: UserManager,
                     private val dao: TreeTrackerDAO,
                     private val workManager: WorkManager,
@@ -86,8 +85,9 @@ class DataViewModel(private val userManager: UserManager,
 
     fun sync() {
         viewModelScope.launch {
-            val treeData = loadPlanterAccountData().treeData
-            val totalCount = treeData.uploadedCount + treeData.waitingToUploadCount
+            val uploadedCount = loadPlanterAccountData().uploadedCount
+            val waitingToUploadCount = loadPlanterAccountData().waitingToUploadCount
+            val totalCount = uploadedCount + waitingToUploadCount
             if (_isSyncing == null || _isSyncing == false) {
                 val treesToSync =
                     withContext(Dispatchers.IO) { dao.getNonUploadedTreeCaptureCount() }
@@ -98,14 +98,14 @@ class DataViewModel(private val userManager: UserManager,
                     isSyncingLiveData.value = true
                     startDataSynchronization()
                 }
-                analytics.syncButtonTapped(totalCount, treeData.uploadedCount, treeData.waitingToUploadCount)
+                analytics.syncButtonTapped(totalCount, uploadedCount, waitingToUploadCount)
             } else {
                 updateTimerJob?.cancel()
                 updateTimerJob = null
                 workManager.cancelUniqueWork(TreeSyncWorker.UNIQUE_WORK_ID)
                 syncNotification.removeNotification()
                 isSyncingLiveData.value = false
-                analytics.stopButtonTapped(totalCount, treeData.uploadedCount, treeData.waitingToUploadCount)
+                analytics.stopButtonTapped(totalCount, uploadedCount, waitingToUploadCount)
             }
         }
     }
@@ -118,11 +118,10 @@ class DataViewModel(private val userManager: UserManager,
             val totalAmountPaid = planterAccountEntity?.totalAmountPaid ?: 0.0
             val paymentAmountPending = planterAccountEntity?.paymentAmountPending ?: 0.0
             val waitingToUploadCount = dao.getNonUploadedTreeImageCount()
-            val treeData = TreeData(uploadedCount = uploadedTreeCount,
-                     waitingToUploadCount = waitingToUploadCount,
-                     validatedCount = validatedTreeCount)
             PlanterAccountData(
-                treeData = treeData,
+                uploadedCount = uploadedTreeCount,
+                waitingToUploadCount = waitingToUploadCount,
+                validatedCount = validatedTreeCount,
                 totalAmountPaid = totalAmountPaid,
                 paymentAmountPending = paymentAmountPending
             )
@@ -131,9 +130,7 @@ class DataViewModel(private val userManager: UserManager,
 
     private fun updateData() {
         viewModelScope.launch(Dispatchers.IO) {
-
             val treeInfo = loadPlanterAccountData()
-
             withContext(Dispatchers.Main) {
                 treeInfoLiveData.value = treeInfo
             }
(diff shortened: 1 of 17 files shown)


app/src/main/res/layout/fragment_data.xml
```diff
@@ -1,132 +1,248 @@
...
+                android:orientation="vertical"
+                app:layout_constraintGuide_begin="243dp" />
+
+            <TextView
+                android:id="@+id/paymentPendingLabel"
+                android:layout_width="0dp"
+                android:layout_height="@dimen/height_forty"
+                android:layout_marginStart="5dp"
+                android:layout_marginTop="16dp"
+                android:layout_weight="3"
```
Elforama: I think all weights can be removed

app/src/main/res/layout/fragment_data.xml
```diff
@@ -1,132 +1,248 @@
...
+                android:id="@+id/paymentDataGuideline"
+                android:layout_width="wrap_content"
+                android:layout_height="wrap_content"
+                android:orientation="vertical"
+                app:layout_constraintGuide_begin="243dp" />
+
+            <TextView
+                android:id="@+id/paymentPendingLabel"
+                android:layout_width="0dp"
+                android:layout_height="@dimen/height_forty"
```
Elforama: It's great that you put this in a style, but can we rename it to something that doesn't mentioned it's value? If we ever change the internal value, we'd need to refactor all usages of the style.

app/src/main/res/layout/fragment_data.xml
```diff
@@ -1,132 +1,248 @@
...
-
-    <LinearLayout
-        android:id="@+id/fragment_data_total_trees"
+<androidx.constraintlayout.widget.ConstraintLayout xmlns:android="http://schemas.android.com/apk/res/android"
+    xmlns:app="http://schemas.android.com/apk/res-auto"
+    xmlns:tools="http://schemas.android.com/tools"
+    android:id="@+id/fragment_data"
+    android:layout_width="match_parent"
+    android:layout_height="match_parent"
+    android:background="@color/lightGrey"
```
Elforama: Can we please have the background here match the color of the login screens?

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/DataViewModel.kt
```diff
@@ -84,45 +86,53 @@ class DataViewModel(private val dao: TreeTrackerDAO,
...
-                analytics.stopButtonTapped(total, synced, waiting)
+                isSyncingLiveData.value = false
+                analytics.stopButtonTapped(totalCount, treeData.uploadedCount, treeData.waitingToUploadCount)
             }
         }
     }
 
-
-    private suspend fun loadTreeInfo(): TreeData {
+    private suspend fun loadPlanterAccountData(): PlanterAccountData {
```
Elforama: This function has gotten very large, can we please move this logic into a "usecase" class that returns the PlanterAccountData?

app/src/main/res/layout/fragment_data.xml
```diff
@@ -1,132 +1,248 @@
...
+                app:layout_constraintTop_toBottomOf="@+id/fragmentDataWaitingUploadLabel" />
+
+            <TextView
+                android:id="@+id/fragmentDataValidatedTrees"
+                android:layout_width="0dp"
+                android:layout_height="@dimen/height_forty"
+                android:layout_marginTop="16dp"
+                android:layout_weight="3"
+                android:textColor="@color/black"
+                android:textIsSelectable="true"
```
Elforama: A lot of the text in this view is selectable, why is that?

app/src/main/res/layout/fragment_data.xml
```diff
@@ -1,132 +1,248 @@
...
+                android:layout_height="@dimen/height_forty"
+                android:layout_marginTop="16dp"
+                android:layout_weight="3"
+                android:text="@string/fragment_data_waiting_to_upload_label"
+                android:textColor="@color/black"
+                android:textSize="@dimen/text_size_seventeen"
+                app:layout_constraintStart_toStartOf="@+id/fragmentDataUploadedLabel"
+                app:layout_constraintTop_toBottomOf="@+id/fragmentDataUploadedLabel" />
+
+            <TextView
```
Elforama: For all the textviews that need to be populated by data, can we please add debug text to it?
e.g. `tools:text="$12.45"`

Elforama:
Can we please add currency symbols to anything that displays currency? @deepwinter What type of currency are we showing?

Elforama:
@arunbakt Overall looks great!

app/src/main/res/layout/fragment_data.xml
arunbakt: Are you referring to ```@dimen/height_forty```?  Do you want to rename it to something like ```@dimen/labelTextHeight``` ?

app/src/main/res/layout/fragment_data.xml
arunbakt: I was trying to match the UI on the issue description. Changed it to the same color as used in login screen now.

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/DataViewModel.kt
arunbakt: The function isn't as big compared to sync function in this class. Since this function is simply accessing data and instantiating a PlanterAccountData it could be part of a Repository so that it could be used by multiple view models if there is a need for it. Right now there is not any usage of that sort I believe.

It feels to me that this method has no sufficient business logic/use case to deserve a use case.  For my clarification, Use case classes I assume are orchestraters of other domain classes to effect a use case, correct?

app/src/main/res/layout/fragment_data.xml
arunbakt: Not sure how this got in, must be my IDE doing it. Removing all of these from this layout

app/src/main/res/layout/fragment_data.xml
arunbakt: Adding them.

arunbakt:
Hi Jonathan,

 I am assuming we are going to use just USD for now for the total paid and pending payment values. I am not accounting for potential multi-currency payments right now. If this is going to be a need, it would be better to have the ability to store the target currency the payment is made in for a planter. For this, we might need to revisit this piece later.

arunbakt:
Thank you for the feedback Jonathan! Updated the commit with changes recommended except for the suggestion to move the planter account data creation to a separate use case. Left the reasoning on the comment sections.

Defaulting to USD for now for the payment values. Will modify if necessary to store the target currency along with planter account data when the server api has clarity on this.

Please let me know what you think.

Thank you,
Arun

Elforama:
This looks good, thanks! We'll just waiting on having this merged since it requires work in other areas (API and stuff)

PR #430, opened 2020-03-22
Add migration for bulk uploads

Elforama:


Elforama:
@arunbakt What is the purpose for the index creation and copying the table? I see that bulk upload ID was added to another table.

arunbakt:
I went by the diff between last two commits on the master branch. There were two changes to the database entities. 
1. That changes `local_photo_path` non-null column to a nullable column in `planter_check_in` table
2. Adding column  `bundle_id` to `planter_info` table

To  migration addressing both these changes to the database schema. The index creation is to be consistent with what existed in the planter_check_in prior to the migration.

PR #433, opened 2020-03-29
417 Planter account payment data integration

arunbakt:
This change includes the api integration (yet to be developed) to retrieve payment data for the planter account and saves it to the local database using the planter identifier. The planter identifier is either the email or phone number associated with the planter.  Each sync invokes the planter account api for all the planters on the device and caches the result on a local db. 

Note: Currently the dev build.gradle points to a locally running nodejs server with a mock implementation for the planter account api and this has to be replaced with the real one when it becomes available.

Test:
1. Install a fresh greenstand treetracker dev variant after removing any local instances already running using ` .\adb uninstall org.greenstand.android.TreeTracker.dev`
2. The migration included will work for any prior release version of the app and thus uninstalling app cleanly is not necessary in the production release version.
3. Run the app and go to tree data fragment and click Sync button. You will see the payment and tree data updated assuming if you have a fake api running locally in your network where the app is running.


app/build.gradle
```diff
@@ -143,6 +143,7 @@ android {
             resValue "string", "app_name", "Tree Tracker(dev)"
             buildConfigField "Boolean", "GPS_ACCURACY", "false"
             buildConfigField 'String', 'BASE_URL', '"http://dev.treetracker.org/api/mobile/"'
+            buildConfigField 'String', 'PLANTER_ACCOUNT_URL', '"http://192.168.2.128:3000/"'
```
arunbakt: Please note, this is a dev url used to test the changes with a locally running planter account api. This needs to be updated when a dev version of the account payment service is available

app/src/main/java/org/greenstand/android/TreeTracker/api/ApiService.kt
```diff
@@ -27,9 +26,15 @@ interface ApiService {
...
 
+    @POST("planter_accounts")
+    suspend fun planterAccountData(
+        @Body planterAccountRequest: PlanterAccountRequest
+    ): Set<PlanterAccountData>
+
     companion object {
 
         const val ENDPOINT = BuildConfig.BASE_URL
+        const val PAYMENT_ENDPOINT = BuildConfig.PLANTER_ACCOUNT_URL
```
arunbakt: When we have one unified api, we could remove `PAYMENT_ENDPOINT`. For now, I am assuming two separate endpoints.

app/src/main/java/org/greenstand/android/TreeTracker/di/NetworkModule.kt
```diff
@@ -41,4 +42,17 @@ val networkModule = module {
             .create(ApiService::class.java)
     }
 
+    single(named("paymentApi")) {
```
arunbakt: we probably may not need the qualifier when we have the final unified api that includes the payment data. For now, my assumption requires one existing endpoint to support existing integrations and the new planter account api for the payment and tree data.

app/build.gradle
```diff
@@ -263,6 +264,7 @@ dependencies {
     api 'com.google.firebase:firebase-core:16.0.6'
 
     implementation "androidx.legacy:legacy-support-v4:${androidSupportVersion}"
+    implementation 'com.android.support:cardview-v7:28.0.0'
```
Elforama: Please refer the androidX version of this library (or the material design library)

app/src/main/java/org/greenstand/android/TreeTracker/api/ApiService.kt
```diff
@@ -27,9 +26,15 @@ interface ApiService {
     @PUT("devices/")
     suspend fun updateDevice(@Body deviceRequest: DeviceRequest): PostResult
 
+    @POST("planter_accounts")
+    suspend fun planterAccountData(
+        @Body planterAccountRequest: PlanterAccountRequest
```
Elforama: Can we please have this on a single line like the other APIs please?

app/src/main/java/org/greenstand/android/TreeTracker/api/models/responses/ResponseTypes.kt
```diff
@@ -0,0 +1,33 @@
+package org.greenstand.android.TreeTracker.api.models.responses
+
+import com.google.gson.annotations.SerializedName
+
+data class UserTree(
```
Elforama: Can we lease have the word Reponse at the end of these data classes so ppl know they are from API?

app/src/main/java/org/greenstand/android/TreeTracker/di/NetworkModule.kt
```diff
@@ -41,4 +42,17 @@ val networkModule = module {
...
+        Retrofit.Builder()
```
Elforama: Can we try to reuse the http client we have for the other API calls? Also, let's try to keep these Module methods void of logic

app/src/main/java/org/greenstand/android/TreeTracker/usecases/GetPlanterAccountData.kt
```diff
@@ -0,0 +1,36 @@
+package org.greenstand.android.TreeTracker.usecases
+
+import org.greenstand.android.TreeTracker.api.RetrofitApi
+import org.greenstand.android.TreeTracker.database.TreeTrackerDAO
+import org.greenstand.android.TreeTracker.database.entity.PlanterAccountEntity
+import timber.log.Timber
+import java.util.*
+
+const val TAG: String = "GetPlanterAccountData"
```
Elforama: This needs be moved into the GetPlanterAccountData class.  Otherwise you'd have multiple top level TAG values in the future.

PR #441, opened 2020-05-05
Location data capture for tree location analysis 

arunbakt:
This is a work in progress and created to receive comments, feedback and discussion. 

The change is to capture location data from the moment the user intends to capture a planted tree until the time it is clicked. If the user decides to cancel, the location capture stops. A shared view model is used by MapsFragment and NewTreeFragment to observe the location data. The shared view model saves the new location to a local db for later syncing it to the cloud for accuracy analysis. This sync part is not yet implemented and is pending.

Please see comments on the code for review feedback and discussion.


app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/CaptureLocationViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+    private val treeTrackerDAO: TreeTrackerDAO,
+    private val workManager: WorkManager
+):ViewModel() {
+
+    private val gson = Gson()
+    private var cancelLocationCaptureJob: CoroutineContext? = null
+    private val locationObserver: Observer<Location?> = Observer {
+        it?.apply {
+            viewModelScope.launch(Dispatchers.IO) {
+                userManager.planterCheckinId?.let{ planterCheckinId ->
```
Elforama, line 33: Can we please move this logic into a useCase class?

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/CaptureLocationViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+                            longitude,
+                            accuracy,
+                            System.currentTimeMillis()
+                        )
+                    val base64String = Base64.encodeToString(
+                        gson.toJson(locationData).toByteArray(),
+                        Base64.DEFAULT
+                    )
+                    Timber.i("Inserting a new location data $base64String")
+                    treeTrackerDAO.insertLocationData(LocationCaptureEntity(base64String))
```
Elforama, line 47: Can we move this logic into a usecase class?

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/CaptureLocationViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+    }
+
+    fun stopLocationCapture() {
+        cancelLocationCaptureJob?.cancel()
+        userLocationManager.locationUpdateLiveDate.removeObserver(locationObserver)
+    }
+
+    private suspend fun scheduleCancelLocationCapture() {
+        viewModelScope.launch(Dispatchers.Main) {
+            delay(1000*60*15)
```
Elforama, line 65: Can we please use constants here to help make more sense of these values

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/CaptureLocationViewModel.kt
```diff
@@ -0,0 +1,78 @@
...
+import kotlinx.coroutines.delay
+import kotlinx.coroutines.launch
+import org.greenstand.android.TreeTracker.database.TreeTrackerDAO
+import org.greenstand.android.TreeTracker.database.entity.LocationCaptureEntity
+import org.greenstand.android.TreeTracker.managers.UserLocationManager
+import org.greenstand.android.TreeTracker.managers.UserManager
+import timber.log.Timber
+import kotlin.coroutines.CoroutineContext
+
+class CaptureLocationViewModel(
```
Elforama, line 21: It seems like we're using this viewmodel less like a viewmodel, and more like a model layer objected with a limited scope.
The two views that use this ViewModel already have Viewmodels, which can get confusing. Can we look into using Koins scoping feature? Perhaps we can create a tree capture scope to help limit the capture time of location.

Elforama:
Looking really good!

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/CaptureLocationViewModel.kt
Elforama: Also, in reference to the Koin scoping below, perhaps the usecase can be what is scoped to the other two views ViewModels

arunbakt:
Thank you for your quick review @Elforama.  I will look into the koin scoped use case tomorrow. 
 I have couple of questions.  
1. I used ViewModel so that I can tap into the lifecycle component feature where the  observer is removed the moment ViewModel scope goes away and this would make it unnecessary to schedule cancellation code using co-routines. The cancellation co-routine uses  Main dispatcher and i haven't tested it. It might not be good thing to use Main thread I suppose. What is your thoughts on it? 
Thank you again for your inputs. It helps to have a conversation.

app/src/main/java/org/greenstand/android/TreeTracker/database/TreeTrackerDAO.kt
```diff
@@ -3,6 +3,7 @@ package org.greenstand.android.TreeTracker.database
 import androidx.room.*
 import org.greenstand.android.TreeTracker.database.entity.*
 import org.greenstand.android.TreeTracker.database.views.TreeMapMarkerDbView
+import org.greenstand.android.TreeTracker.viewmodels.LocationData
```
Elforama, line 6: Do we need this?

app/src/main/java/org/greenstand/android/TreeTracker/usecases/UploadLocationData.kt
```diff
@@ -0,0 +1,44 @@
...
+    override suspend fun execute(params: Unit): Boolean {
+        try {
+            withContext(Dispatchers.IO) {
+                val locations = dao.getTreeLocationData()
+                val treeLocations = mutableListOf<String>()
+                val gson = Gson()
+                for (location in locations) {
+                    val locJson = gson.toJson(location)
+                    treeLocations.add(locJson)
+                }
```
Elforama, line 27: We can make this:
`
val treeLocations = locations.map { gson.toJson(it) } 
`

arunbakt:
@Elforama  I pinged you in slack with a question about using Koin scoped components. I also left comments/questions on this PR regarding the same (I couldn't respond individually to the comments for some reason).  Let me know what you think when you have a chance.

arunbakt:
@Elforama , 

Updated the PR with the suggestions you made. This update includes the sync to upload the file with base64 encoded json data to the s3 account. The only thing i didn't pursue is the scheduled cancellation of data capture in case when the user abandons the capture process for more than 15 minutes. I believe the `onPause` lifecycle method should take care of this if the user navigates away from the capture screen.  

Thank you for your feedback, it was helpful. I like the koin scope better than view model that I initially used.

app/src/main/java/org/greenstand/android/TreeTracker/usecases/SyncDataUseCase.kt
```diff
@@ -8,9 +8,12 @@ import org.greenstand.android.TreeTracker.database.TreeTrackerDAO
 import timber.log.Timber
 import kotlin.coroutines.coroutineContext
 
+const val TAG = "SyncDataUseCase"
```
Elforama, line 11: Can we move this into the Usecase? Otherwise it's a global value

app/src/main/java/org/greenstand/android/TreeTracker/usecases/SyncDataUseCase.kt
```diff
@@ -70,6 +74,13 @@ class SyncDataUseCase(
     }
 
     private suspend fun uploadTreeLocationData() {
-        // Timber.tag("UploadTreeLocationData").d("Uploading ${}")
+        Timber.tag(TAG).d("Processing tree location data")
+        kotlin.runCatching {
+            uploadLocationDataUseCase.execute(Unit)
+            Timber.tag(TAG).d("Uploading tree location data complete")
+            dao.purgeUploadedTreeLocations()
```
Elforama, line 81: What do you think about moving this purge into the upload usecase you made?

Elforama:
@arunbakt Looking good!

I have a question about when we are supposed to capture the users location. It looks like we captures it for as long as the camera is shown to the user. Is it shown more than that?

app/src/main/java/org/greenstand/android/TreeTracker/activities/ImageCaptureActivity.kt
```diff
@@ -54,6 +61,16 @@ class ImageCaptureActivity : AppCompatActivity() {
         viewFinder.post { startCamera(captureSelfie) }
     }
 
+    override fun onPause() {
+        super.onPause()
+        locationCaptureScope.stopLocationCapture()
```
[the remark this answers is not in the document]
Elforama: If the GPS is only tracking while they are in the camera, why don't we start GPS when the camera screen starts, and stop it when the camera screen starts. Not sure we need the start GPS in maps fragment or if we need the custom scope either.

app/src/main/java/org/greenstand/android/TreeTracker/activities/ImageCaptureActivity.kt
```diff
@@ -54,6 +61,16 @@ class ImageCaptureActivity : AppCompatActivity() {
...
+    }
+
+    override fun onResume() {
+        super.onResume()
+        locationCaptureScope.startLocationCapture()
```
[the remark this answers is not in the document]
Elforama: We might want to call these in onStart and onStop instead. onPause can get called if a dialog pops up.
I also think we may want to limit where the GPS logic is, and just keep in in the camera screen. For instance, if we put a screen before the camera, we could have GPS running when it shouldn't.

app/src/main/java/org/greenstand/android/TreeTracker/activities/ImageCaptureActivity.kt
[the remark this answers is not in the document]
arunbakt: @Elforama,  you have a valid point. But the original requirement was to expand the GPS data collection from the moment user clicks on Add tree button.  I agree, if the need is to just capture  location data only when the camera is shown to the user, we don't need the scoped component and limit it to the camera screen alone. I will update the PR.

arunbakt:
@Elforama updated the PR with your suggestion to restrict the location data collection to the camera screen.  Please take a look when you have a chance.

Issue #436, opened 2020-04-22
Add UUID to planter registration JSON


deepwinter:
Add a record_uuid to the planter registration JSON.  This UUID is only for detecting duplicate inserts later in the pipeline, it has no significance in identifying the planter.

{ lat: 0,
  long: 0,
  last_name: 'uh',
  first_name: 'hh',
  organization: 'uj',
  device_identifier: 'ab0045b962db2543',
  planter_identifier: '3126655658',
  record_uuid: $calculated_uuid }

PR #442, opened 2020-05-15
436 record UUID

Elforama:
Issue: https://github.com/Greenstand/treetracker-android/issues/436

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/database/Migrations.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/database/Migrations.kt
@@ -2,6 +2,7 @@ package org.greenstand.android.TreeTracker.database
 
 import androidx.room.migration.Migration
 import androidx.sqlite.db.SupportSQLiteDatabase
+import org.greenstand.android.TreeTracker.database.entity.PlanterInfoEntity
 
 
 val MIGRATION_1_2 = object : Migration(1, 2) {
@@ -55,4 +56,10 @@ val MIGRATION_1_2 = object : Migration(1, 2) {
         }
     }
 
+}
+
+val MIGRATION_2_3 = object : Migration(2, 3) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL("ALTER TABLE ${PlanterInfoEntity.TABLE} ADD COLUMN ${PlanterInfoEntity.RECORD_UUID} TEXT")
+    }
 }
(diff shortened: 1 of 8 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/database/entity/PlanterInfoEntity.kt
```diff
@@ -27,7 +27,9 @@ data class PlanterInfoEntity(
     @ColumnInfo(name = CREATED_AT)
     var createdAt: Long,
     @ColumnInfo(name = BUNDLE_ID)
-    var bundleId: String? = null
+    var bundleId: String? = null,
+    @ColumnInfo(name = RECORD_UUID)
+    var recordUuid: String? = null
```
arunbakt, line 32: When would the `recordUuid` will be null?  Won't it be a non-null value always?

app/src/main/java/org/greenstand/android/TreeTracker/database/Migrations.kt
```diff
@@ -55,4 +56,10 @@ val MIGRATION_1_2 = object : Migration(1, 2) {
         }
     }
 
+}
+
+val MIGRATION_2_3 = object : Migration(2, 3) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        database.execSQL("ALTER TABLE ${PlanterInfoEntity.TABLE} ADD COLUMN ${PlanterInfoEntity.RECORD_UUID} TEXT")
```
arunbakt, line 63: Shouln't we require a non-null value for `record_uuid`?

app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
```diff
@@ -236,6 +235,9 @@ class MapsFragment : androidx.fragment.app.Fragment(), OnClickListener, OnMapRea
     }
 
     private fun renderTrees() {
+
+        map ?: return
```
arunbakt, line 239: In what conditions would the map be null? When location is not enabled?

Elforama: While working on this it crashed here due to it being null. This is a quick fix

app/src/main/java/org/greenstand/android/TreeTracker/database/Migrations.kt
Elforama: Update

app/src/main/java/org/greenstand/android/TreeTracker/database/entity/PlanterInfoEntity.kt
Elforama: Updated to non null

PR #450, opened 2020-07-10
Update travis config to make PR builds not depend on decrypting property files

arunbakt:
Modified travis file to conditionally un-encrypt key values so that PR builds doesn't require the encryption key values that test and prod builds depend upon.

This is WIP and have to be tested in the build pipeline.


Elforama:
Not sure what to make of this. But looks good. Never done travis before. Was this tested?

arunbakt:
This is to eliminate the execution of the following command that fails on running  PR builds
``` openssl aes-256-cbc -K $encrypted_0a37c10e9e19_key -iv $encrypted_0a37c10e9e19_iv
  -in secrets.tar.enc -out secrets.tar -d```

The reason for the failure is missing values for `encrypted_0a37c10e9e19_key` and `encrypted_0a37c10e9e19_iv`. 
I assume we do this to avoid disclosing our test env creds to the public but are required as per travis config.

The change conditionally omits this for PR builds and instead uses a dummy treetracker.keys.properties to have them run.
Other change is to use `gradlew build` to execute unit tests when we add them  instead of the `assemblyGreenstandDebug` currently used. 

I tested in my local machine but this has to tested in the travis environment . Also, I noticed @jhoon had some changes to address the same issue yesterday but I don't see those changes (possibly he/she reverted it). So it would help if @jhoon could also review this change.

jhoon:
Hey @arunbakt yeah, I had changes on #419 which is correctly passing Travis CI, though I was wondering if anyone took a look at it already. @Elforama, @deepwinter or anyone else maybe could? -it's been there a while

jhoon:
Solved with #419

PR #453, opened 2020-07-20
Disregard automatic sign out when adding a tree in justdigit version

arunbakt:
The justdigit version of treetracker requires automatic sign out to not occur. Currently the default version has a background job that signs out the user if the last check in is two weeks old. But there is also another check in feature when the user is attempting to add a tree. This check makes check-in older than 2 hours requiring a login again. 

The solution is add the automatic sign-out feature flag to this condition thereby just-digit versions don't require this check while adding a tree.

### DIFF
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -36,7 +36,7 @@
         android:name=".application.TreeTrackerApplication"
         android:allowBackup="true"
         android:icon="@mipmap/ic_launcher"
-        android:roundIcon="@mipmap/ic_launcher"
+        android:roundIcon="@mipmap/ic_launcher_round"
         android:supportsRtl="true"
         android:label="@string/app_name"
         android:theme="@style/AppTheme"
--- a/app/src/main/AndroidManifest.xml
+++ b/app/src/main/AndroidManifest.xml
@@ -36,7 +36,7 @@
         android:name=".application.TreeTrackerApplication"
         android:allowBackup="true"
         android:icon="@mipmap/ic_launcher"
-        android:roundIcon="@mipmap/ic_launcher_round"
+        android:roundIcon="@mipmap/ic_launcher"
         android:supportsRtl="true"
         android:label="@string/app_name"
         android:theme="@style/AppTheme"
--- a/app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
@@ -40,7 +40,6 @@ import org.greenstand.android.TreeTracker.managers.FeatureFlags
 import org.greenstand.android.TreeTracker.managers.UserLocationManager
 import org.greenstand.android.TreeTracker.managers.accuracyStatus
 import org.greenstand.android.TreeTracker.map.TreeMapMarker
-import org.greenstand.android.TreeTracker.usecases.ExpireCheckInStatusUseCase
 import org.greenstand.android.TreeTracker.utilities.ImageUtils
 import org.greenstand.android.TreeTracker.utilities.TreeClusterRenderer
 import org.greenstand.android.TreeTracker.utilities.ValueHelper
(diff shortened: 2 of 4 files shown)


app/src/main/AndroidManifest.xml
```diff
@@ -36,7 +36,7 @@
         android:name=".application.TreeTrackerApplication"
         android:allowBackup="true"
         android:icon="@mipmap/ic_launcher"
-        android:roundIcon="@mipmap/ic_launcher_round"
+        android:roundIcon="@mipmap/ic_launcher"
```
Elforama, line 39: Why was the icon shape here changed?

app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
```diff
@@ -164,7 +164,8 @@ class MapsFragment : androidx.fragment.app.Fragment(), OnClickListener, OnMapRea
                         ValueHelper.TIME_OF_LAST_PLANTER_CHECK_IN_SECONDS,
                         0
                     )
-                    if (currentTimestamp - lastTimeStamp > ValueHelper.CHECK_IN_TIMEOUT) {
+                    if (FeatureFlags.AUTOMATIC_SIGN_OUT_FEATURE_ENABLED &&
```
Elforama, line 167: We should replace all this logic with `ValidateCheckInStatusUseCase`

app/src/main/AndroidManifest.xml
arunbakt: For some reason, I don't have the ic_launcher_round in my checkout files. I will double check this again.

app/src/main/java/org/greenstand/android/TreeTracker/usecases/ExpireCheckInStatusUseCase.kt
```diff
@@ -14,4 +17,13 @@ class ExpireCheckInStatusUseCase constructor(private val sharedPreferences: Shar
...
+    fun readyForAutomaticSignout(): Boolean {
+        val currentTimestamp = System.currentTimeMillis() / 1000
+        val timeStampForLastCheckIn = sharedPreferences.getLong(
+            ValueHelper.TIME_OF_LAST_PLANTER_CHECK_IN_SECONDS,
+            0
+        )
+        return (FeatureFlags.AUTOMATIC_SIGN_OUT_FEATURE_ENABLED &&
+                currentTimestamp - timeStampForLastCheckIn > ValueHelper.CHECK_IN_TIMEOUT)
+    }
+}
```
Elforama, line 29: Ideally we want a usecase to have just the execute method. It should do just one thing.
Also, as I mentioned in a past comment, this logic is already in the app inside `ValidateCheckInStatusUseCase` we just need to trigger that where the check is. Ideally from within the viewmodel

arunbakt: I will make the change. I didn't notice the `ValidateCheckInStatusUseCase`.

 I prefer and like domain models that have cohesive responsibilities and not thrilled by functions dispersed in various disparate use cases. I understand this is how the app is architected. In my mind, UseCase pattern feels appropriate when there is a notion of performing an action that requires orchestrating more than one domain class. If we were to create use case for every single piece of logic than there is no cohesive domain object.  Just wanted to share my thoughts with you.

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/MapViewModel.kt
```diff
@@ -33,6 +35,10 @@ class MapViewModel constructor(private val validateCheckInStatusUseCase: Validat
         }
     }
 
+    suspend fun requiresLogin(): Boolean {
```
arunbakt, line 38: I had to add this function since I don't see a good way to use the existing boolean MutableLiveData in a `Add tree button` click listener. Let me know if there is a clean and simple way to use it, I will change it. 

For now, this looks like the straight forward way to implement this change.

PR #454, opened 2020-07-24
Migration to 1.3

arunbakt:
This addresses the migration of 1.2.7 release database version to 1.3 release version.
It migrates the schema from what existed in 1.2.7 to the 1.3 version while preserving the data. 

Caveat:
The only loss expected is the `image_url` in `tree_capture` which is obtained after syncing the data to the cloud which is set to be saved in 1.3 version but wasn't the case with the older 1.2.7 version. Thus tree synced in 1.2.7 won't have this path and thus won't have any value in its equivalent 1.3 version.

Tests: 

This migration is tested on a clean install of 1.2.7 with few trees captured in an emulator and migrating it to 1.3 and syncing the data. No loss was detected or crash occurred.

Note: While this change is under review, will attempt to see if we can test in on a real device with production data on a individual basis to validate this before rolling it out to all if possible.

Also, this change needs to be merged with the master version to avoid losing this history. I will create a seperate PR for that after this gets merged.


app/src/main/java/org/greenstand/android/TreeTracker/database/AppDatabase.kt
```diff
@@ -34,14 +34,14 @@ abstract class AppDatabase : RoomDatabase() {
...
-                        .addMigrations(MIGRATION_1_2)
+                        .addMigrations(MIGRATION_3_4)
                         .build()
                 }
             }
             return INSTANCE!!
         }
 
-        private const val DB_NAME = "treetracker.v1.db"
+        private const val DB_NAME = "treetracker.v2.db"
```
Elforama, line 44: Why was this changed?

app/src/main/java/org/greenstand/android/TreeTracker/database/Migrations.kt
```diff
@@ -3,49 +3,180 @@ package org.greenstand.android.TreeTracker.database
 import androidx.room.migration.Migration
 import androidx.sqlite.db.SupportSQLiteDatabase
 
-
-val MIGRATION_1_2 = object : Migration(1, 2) {
+val MIGRATION_3_4 = object : Migration(3, 4) {
```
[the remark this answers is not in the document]
arunbakt: Yes, the older version is at 3, hence the migration from 3 to 4

app/src/main/java/org/greenstand/android/TreeTracker/database/AppDatabase.kt
arunbakt: The older db is named `v2.db`. Since we are migrating the db, I believe it is better to keep the name the same. This would keep the variations to a minimum and keep the migration script less complicated.

arunbakt:
I tested the migration in a personal android device (samsung galaxy) by adding trees in 1.2.7 version and upgrading it to 1.3. All the tree data are preserved by the migration.

One thing I couldn't test was doing a partial sync in 1.2.7 and upgrading the app version.  This is because I don't have the correct property files used in 1.2.7 to enable syncing. Any attempt to sync in 1.2.7 version fails.  But once the app is migrated to 1.3, sync starts working obviously because I have the correct property files.

PR #459, opened 2020-07-28
Stop location update listener conditionally

arunbakt:
`UserLocationManager` removes the `LocationUpdateListener`  when the Main activity goes out of scope. This causes the location data collection to not occur when capturing a tree using the ImageCaptureActivity. The location data stream is stopped and hence no updates are posted to the `locationUpdateLiveData` which `CaptureTreeLocationUseCase` depends on.

 The fix/change addresses this by checking if there is an observer for the LocationUpdateLiveData and only when absent to remove the `LocationUpdateListener` from the android's `LocationManager` thereby allowing the location stream to happen when entering `ImageCaptureActivity` by leaving `MainActivity`.

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/managers/UserLocationManager.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/managers/UserLocationManager.kt
@@ -18,7 +18,7 @@ class UserLocationManager(
 ) {
 
     private val locationUpdates = MutableLiveData<Location?>()
-    val locationUpdateLiveDate: LiveData<Location?> = locationUpdates
+    val locationUpdateLiveData: LiveData<Location?> = locationUpdates
 
     var isUpdating: Boolean = false
         private set
@@ -71,7 +71,7 @@ class UserLocationManager(
 
     fun stopLocationUpdates() {
         Timber.d("Request to stop location updates submitted")
-        if (locationUpdateLiveDate.hasObservers())
+        if (locationUpdateLiveData.hasObservers())
             return
         Timber.d("Request to stop location update honored")
         locationManager.removeUpdates(locationUpdateListener)
--- a/app/src/main/java/org/greenstand/android/TreeTracker/managers/UserLocationManager.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/managers/UserLocationManager.kt
@@ -90,14 +90,17 @@ class UserLocationManager(
     }
 
     private fun hasLocationPermissions(): Boolean {
-        return ContextCompat.checkSelfPermission(
-                    context,
-                    android.Manifest.permission.ACCESS_FINE_LOCATION
-                ) == PackageManager.PERMISSION_GRANTED ||
-                ContextCompat.checkSelfPermission(
-                    context,
-                    android.Manifest.permission.ACCESS_COARSE_LOCATION
-                ) == PackageManager.PERMISSION_GRANTED
+
+        val fineLocationPermission = ContextCompat.checkSelfPermission(
+            context,
+            android.Manifest.permission.ACCESS_FINE_LOCATION
+        )
+        val coarseLocationPermission = ContextCompat.checkSelfPermission(
+            context,
+            android.Manifest.permission.ACCESS_COARSE_LOCATION
+        )
+        return (fineLocationPermission == PackageManager.PERMISSION_GRANTED ||
+                coarseLocationPermission == PackageManager.PERMISSION_GRANTED)
     }
 }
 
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/managers/UserLocationManager.kt
```diff
@@ -55,7 +54,12 @@ class UserLocationManager(private val locationManager: LocationManager,
     fun startLocationUpdates(): Boolean {
         return if (hasLocationPermissions()) {
             if (!isUpdating) {
-                locationManager.requestLocationUpdates(LocationManager.GPS_PROVIDER, 0, 0f, locationUpdateListener)
+                locationManager.requestLocationUpdates(
```
arunbakt, line 57: These are ktlint formatting changes.

app/src/main/java/org/greenstand/android/TreeTracker/managers/UserLocationManager.kt
```diff
@@ -55,7 +54,12 @@ class UserLocationManager(private val locationManager: LocationManager,
...
+                    LocationManager.GPS_PROVIDER,
+                    0,
```
arunbakt, line 59: We need to tweak the values. Currently I observe location update every second. This will have some battery consumption. Can we update this to every 5 seconds?

Also, we have this location data stream all the time when any fragments  of Treetracker app is in use. Do we really need it?

app/src/main/java/org/greenstand/android/TreeTracker/managers/UserLocationManager.kt
```diff
@@ -66,6 +70,10 @@ class UserLocationManager(private val locationManager: LocationManager,
     }
 
     fun stopLocationUpdates() {
+        Timber.d("Request to stop location updates submitted")
+        if (locationUpdateLiveDate.hasObservers())
```
arunbakt: This is the main fix I have added to address the location being not captured in the local storage.

app/src/main/java/org/greenstand/android/TreeTracker/managers/UserLocationManager.kt
```diff
@@ -81,13 +89,16 @@ class UserLocationManager(private val locationManager: LocationManager,
         } ?: false
     }
 
-    private fun hasLocationPermissions() : Boolean {
-        return ContextCompat.checkSelfPermission(context,
-                                                 android.Manifest.permission.ACCESS_FINE_LOCATION) == PackageManager.PERMISSION_GRANTED
-                || ContextCompat.checkSelfPermission(context,
-                                                     android.Manifest.permission.ACCESS_COARSE_LOCATION) == PackageManager.PERMISSION_GRANTED
+    private fun hasLocationPermissions(): Boolean {
+        return ContextCompat.checkSelfPermission(
```
Elforama, line 93: I know this is probably for the klint, but I think we should format this ourself. This is very hard to read

app/src/main/java/org/greenstand/android/TreeTracker/managers/UserLocationManager.kt
```diff
@@ -81,13 +89,16 @@ class UserLocationManager(private val locationManager: LocationManager,
...
+                    context,
+                    android.Manifest.permission.ACCESS_FINE_LOCATION
+                ) == PackageManager.PERMISSION_GRANTED ||
+                ContextCompat.checkSelfPermission(
+                    context,
+                    android.Manifest.permission.ACCESS_COARSE_LOCATION
+                ) == PackageManager.PERMISSION_GRANTED
```
Elforama, line 100: I know this is probably for the klint, but I think we should format this ourself. This is very hard to read

app/src/main/java/org/greenstand/android/TreeTracker/managers/UserLocationManager.kt
Elforama: I agree, this can be an issue for battery. If it's every 5 second though, the tree capture data may not be accurate. 

I know we're working on a design that doesn't have the map on the home screen, and that will help a lot in reducing GPS usage once we redesign the app UI.

Elforama:
Does this need to go into the release, or is this just for the current master?

app/src/main/java/org/greenstand/android/TreeTracker/managers/UserLocationManager.kt
arunbakt: Actually, ktlint complained it cannot auto fix these lines. So this is me formatting to meet the line length rule set by ktlint.
I will restructure these lines to simplify them. Thank you.

arunbakt:
> Does this need to go into the release, or is this just for the current master?

@Elforama  I believe 1.3 didn't have location data update, and I think it is being tested by @deepwinter for now. He would know when this is going out.

PR #460, opened 2020-08-03
Bring migration script up-to date in master

arunbakt:
Migration script for upgrading 1.2.7 to 1.3 resulted in version number difference, db package name difference, migration script differences between the production version and what exists in master.

This PR merges the changes made for 1.3 release along with what has been added to master. This will keep the app db consistent with future releases while maintaining history.

### DIFF
--- a/app/schemas/org.greenstand.android.TreeTracker.database.AppDatabase/6.json
+++ b/app/schemas/org.greenstand.android.TreeTracker.database.AppDatabase/6.json
@@ -1,0 +1,422 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 6,
+    "identityHash": "238caf84a0eb4bedccdeb189bf9f7a17",
+    "entities": [
+      {
+        "tableName": "planter_check_in",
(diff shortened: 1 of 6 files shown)


app/schemas/org.greenstand.android.TreeTracker.database.AppDatabase/6.json
```diff
@@ -0,0 +1,422 @@
+{
```
arunbakt, line 1: Ignore this file, automatic generation

app/src/main/java/org/greenstand/android/TreeTracker/database/TreeTrackerDAO.kt
```diff
@@ -34,9 +44,8 @@ interface TreeTrackerDAO {
     @Query("UPDATE planter_info SET uploaded = :isUploaded WHERE _id IN (:ids)")
     fun updatePlanterInfoUploadStatus(ids: List<Long>, isUploaded: Boolean)
 
-
     @Transaction
-    @Query("SELECT * FROM planter_check_in WHERE photo_url IS null AND planter_info_id = :planterInfoId")
+    @Query("SELECT * FROM planter_check_in WHERE photo_url IS null AND planter_info_id = :planterInfoId") // kt-lint-disable max-line-length
```
arunbakt, line 48: For aesthetics and readability I let this Query violate the ktlint's line length. It will be odd to break the single line annotation to two lines.

app/src/main/java/org/greenstand/android/TreeTracker/database/entity/PlanterInfoEntity.kt
```diff
@@ -40,7 +40,7 @@ data class PlanterInfoEntity(
         const val TABLE = "planter_info"
 
         const val ID = "_id"
-        const val IDENTIFIER = "planterInfoId"
+        const val IDENTIFIER = "planter_identifier"
```
arunbakt, line 43: This change is already made in 1.3. I find `planterInfoId` confusing and changed it to `planter_identifier` (this way it is distinct from `_id` that belongs to `planter_info` table).  Also using snake case for the column name.

PR #467, opened 2020-08-19
Location data captured for the duration the app is in foreground

arunbakt:
This change makes location capture to occur for the duration the app is in the foreground.
Left comments on the changes to make sense of changes where necessary.

Tests:
- Tested the location capture and sync with app running in the foreground for an hour and 20 minutes. Resulting in a 5000+ location data that did synced fine.
- The upload of location data requires atleast one tree that is waiting to be synced for the location data to be picked up for syncing.


app/build.gradle
```diff
@@ -53,7 +53,7 @@ android {
         minSdkVersion 21
         targetSdkVersion 28
         versionCode 113
-        versionName "1.2.8"
+        versionName "1.4.0"
```
arunbakt, line 56: This change is targeted for 1.4.0 release. Hence the change. (also we are already at 1.3.0 in production I believe)

app/src/main/java/org/greenstand/android/TreeTracker/activities/ImageCaptureActivity.kt
```diff
@@ -16,20 +16,17 @@ import androidx.camera.core.Preview
...
 import org.greenstand.android.TreeTracker.viewmodels.NewTreeViewModel.Companion.FOCUS_THRESHOLD
-import org.koin.android.ext.android.inject
 import timber.log.Timber
 
 class ImageCaptureActivity : AppCompatActivity() {
 
     private lateinit var viewFinder: TextureView
     private lateinit var imageCaptureButton: ImageButton
     private lateinit var toolbarTitle: TextView
-    private val captureTreeLocationUseCase: CaptureTreeLocationUseCase by inject()
```
arunbakt, line 32: These objects are not necessary for now. But they may be introduced in subsequent PRs when adding tree-uuids. This change is specific to this PR given the requirement.

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
```diff
@@ -18,25 +18,35 @@ import androidx.core.content.ContextCompat
...
 import org.greenstand.android.TreeTracker.utilities.ValueHelper
+import org.koin.android.ext.android.get
 import org.koin.android.ext.android.inject
 
 class MainActivity : AppCompatActivity(), ActivityCompat.OnRequestPermissionsResultCallback {
 
     private val languageSwitcher: LanguageSwitcher by inject()
     private val userManager: UserManager by inject()
     private val analytics: Analytics by inject()
-    private val userLocationManager: UserLocationManager by inject()
```
arunbakt, line 39: Renamed `UserLocationManager` to `LocationUpdateManager` given there are not userId or planterId concerns in this class. Given in the new context, all that `LocationUpdateManager` is concerned with and responsible for is managing when to start, stop and post changes in Location using android's location stream via its listener.

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
```diff
@@ -53,8 +63,10 @@ class MainActivity : AppCompatActivity(), ActivityCompat.OnRequestPermissionsRes
 
         if (sharedPreferences.getBoolean(ValueHelper.FIRST_RUN, true)) {
 
-            if (sharedPreferences.getBoolean(ValueHelper.TREE_TRACKER_SETTINGS_USED, true)) {
-                sharedPreferences.edit()?.putBoolean(ValueHelper.TREE_TRACKER_SETTINGS_USED, true)?.apply()
+            if (sharedPreferences.getBoolean(
```
arunbakt, line 66: These are manual formatting changes to adhere to the line length constraints set by ktlint. This is not auto-formatted and I have used my best judgment to make this readable while being within the line length limits.

app/src/main/java/org/greenstand/android/TreeTracker/di/AppModule.kt
```diff
@@ -56,18 +82,18 @@ val appModule = module {
...
 
     single { NotificationManagerCompat.from(get()) }
 
     single {
-        CaptureTreeLocationUseCase(
-                get(),
-                get(),
-                get()
-            )
+        LocationDataCapturer(
```
arunbakt, line 92: Renamed the `CaptureTreeLocationUseCase` to `LocationDataCapturer` and moved it out of the usecases package and into LocationUpdateManager.  It wasn't inheriting from the `UseCase` class anyways and I don't feel like it is a usecase.
Like in my other observations, `UseCase`  for me is a coordination of objects to perform a process or function at a higher level.

app/src/main/java/org/greenstand/android/TreeTracker/managers/LocationUpdateManager.kt
```diff
@@ -120,3 +129,51 @@ fun Location?.accuracyStatus(): Accuracy {
...
+                    val base64String = Base64.encodeToString(
+                        gson.toJson(locationData).toByteArray(),
+                        Base64.NO_WRAP
+                    )
+                    Timber.d("Inserting a new location data $base64String")
+                    treeTrackerDAO.insertLocationData(LocationDataEntity(base64String))
+                }
+            }
+        }
+        locationUpdateManager.locationUpdateLiveData.observeForever(locationObserver)
```
Elforama, line 161: Why was the start/stop methods removed? It now starts automatically.

app/src/main/java/org/greenstand/android/TreeTracker/managers/LocationUpdateManager.kt
```diff
@@ -120,3 +129,51 @@ fun Location?.accuracyStatus(): Accuracy {
         Accuracy.BAD
     }
 }
+
+class LocationDataCapturer(
```
Elforama, line 133: Can this be it's own file?

arunbakt: For now,  I like to use the kotlin's flexibility to use one file to contain related classes. If the LocationDataCapturer becomes big to deserve its own file, we could create one.

app/src/main/java/org/greenstand/android/TreeTracker/managers/LocationUpdateManager.kt
arunbakt: Yes, the moment the class is instantiated, it starts listening to the updates from LocationUpdateManager.

app/src/main/java/org/greenstand/android/TreeTracker/managers/LocationUpdateManager.kt
```diff
@@ -120,3 +129,51 @@ fun Location?.accuracyStatus(): Accuracy {
...
+    private val userManager: UserManager,
+    private val locationUpdateManager: LocationUpdateManager,
+    private val treeTrackerDAO: TreeTrackerDAO
+) : KoinComponent {
```
Elforama, line 137: Why is this a KoinComponent?

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
```diff
@@ -18,25 +18,35 @@ import androidx.core.content.ContextCompat
...
     private val userManager: UserManager by inject()
     private val analytics: Analytics by inject()
-    private val userLocationManager: UserLocationManager by inject()
+    private val locationUpdateManager: LocationUpdateManager by inject()
+
+    // An eager instance of LocationDataCapturer assures the location data capture for the duration
+    // the Treetracker App is in the foreground as required for data analysis. Even if this instance
+    // is not used explicitly in this activity, this declaration is necessary to initiate the
+    // process.
+    private val locationDataCapturer: LocationDataCapturer = get()
```
Elforama, line 48: Why not just inject lazily and call the start method?

app/src/main/java/org/greenstand/android/TreeTracker/managers/LocationUpdateManager.kt
arunbakt: It allows to use the `get()` making an eager instantiation possible.

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
arunbakt: I considered it as well. Will change this to have a start method, I am ok with that.
Just curious, is there any reason to not use KoinComponent other than the preference for a simple pojo?

arunbakt: Updated the PR with this change, this choice is more explicit. Thank you for the feedback @Elforama

app/src/main/java/org/greenstand/android/TreeTracker/managers/LocationUpdateManager.kt
Elforama: This ties our code to Koin, so I think it's best to avoid it. If the only reason is eager instantiation, then we probably don't need it. Ideally everything should be lazy until we need it.

PR #469, opened 2020-08-20
Tag all location data with a UUID and associate it with the tree captured

arunbakt:
In add tree mode, all location data are tagged with UUID which is then used as the tree's uuid when the planter saves the captured tree. 

When planter clicks on 'Add tree' button, the locations captured will have a UUID tagged to them. When planter completes clicking the camera button, the tagging of UUID with location data stops. Subsequently, if the planter saves the captured picture of the tree, the UUID  tagged in the locations will be used  to create the entry for the captured tree.

If the user abandons without saving the captured picture, the locations tagged will persist. But no entry for the tree will be made as expected.


app/src/main/java/org/greenstand/android/TreeTracker/fragments/NewTreeFragment.kt
```diff
@@ -111,10 +116,15 @@ class NewTreeFragment :
...
         } else {
             findNavController().popBackStack()
+            locationDataCapturer.turnOffTreeCaptureMode()
         }
     }
 
     override fun onActivityResult(requestCode: Int, resultCode: Int, data: Intent?) {
+
+        newTreeUuid = locationDataCapturer.generatedTreeUuid
+        locationDataCapturer.turnOffTreeCaptureMode()
```
arunbakt, line 126: We turn off the tree capture mode immediately on a activity result from the Camera screen. This is as per the spec as well to stop the ongoing Locations that are captured from getting tagged with the tree uuid generated in the prior 'Add Tree' flow. The cached `newTreeUuid` is retained and used to create the tree captured entity if the planter proceeds to save it.

Elforama:
I think you need to rebase this. It has changes from the previous PR.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/NewTreeFragment.kt
```diff
@@ -13,25 +13,30 @@ import androidx.core.app.ActivityCompat
...
+import org.koin.android.ext.android.inject
 import org.koin.android.viewmodel.ext.android.viewModel
 import timber.log.Timber
 
 class NewTreeFragment :
     androidx.fragment.app.Fragment(), ActivityCompat.OnRequestPermissionsResultCallback {
 
     private val vm: NewTreeViewModel by viewModel()
+    private val locationDataCapturer by inject<LocationDataCapturer>()
+    private var newTreeUuid: UUID? = null
```
Elforama, line 39: Can we move these two fields to the viewmodel?

When they are in the view like this it makes it harder to test later on. Ideally the view just renders stuff.

Elforama:
In the case of the pending PR, there aren't many changes happening at once to this repo. So it's probably safe to just keep rebasing. I wouldn't flatten the commits since they are supposed to be two PRs.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/NewTreeFragment.kt
arunbakt: `locationDataCapturer` is in itself a model. It makes sense to move these behind view model. But in the case of `mapsFragment` the context doesn't have a natural flow. I will make the change and you might see what I am talking about.

arunbakt:
@Elforama , I updated this PR as per your feedback, please take a look when you get a chance.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
```diff
@@ -163,6 +165,7 @@ class MapsFragment : androidx.fragment.app.Fragment(), OnClickListener, OnMapRea
                             findNavController().navigate(
                                 MapsFragmentDirections.actionGlobalLoginFlowGraph())
                         } else {
+                            vm.turnOnTreeCaptureMode()
```
arunbakt, line 168: This is the change I feel is kind of weird. Invoking `turnOnTreeCaptureMode` on `locationDataCapturer` is much more clearer in terms of the flow than invoking it in the view model (which in turn invokes the method with the same name on `locationDataCapturer`). But I understand mvvm pattern recommends no direct interaction with other types of models on the UI code.

Issue #476, opened 2020-08-29
Resize images manually after capture


Elforama:
We are requesting images of the size 800x800. However, some phones do not support this and end up capture very high resolution images instead. In the case that the captured image is not the right size we want to resize it manually and re-save it.

PR #477, opened 2020-08-30
476 scale image

Elforama:
https://github.com/Greenstand/treetracker-android/issues/476


app/src/main/java/org/greenstand/android/TreeTracker/utilities/ImageUtils.kt
```diff
@@ -297,38 +296,134 @@ object ImageUtils {
...
+
+        val compressionQuality = 100
+        val byteArrayBitmapStream = ByteArrayOutputStream()
+        rotatedBitmap.compress(Bitmap.CompressFormat.PNG, compressionQuality,
+            byteArrayBitmapStream)
+
+        return rotatedBitmap
+    }
+
+    fun base64Image(path: String): String {
```
arunbakt, line 380: I see this function is not used anywhere, why do we need it?  Also, if there is a need, can't this function just call `resizedImage` and then proceed to call `Base64.encodeToString()` instead of implementing the same code as `resizedImage()` does? Or am I missing something nuance.

app/src/main/java/org/greenstand/android/TreeTracker/utilities/ImageUtils.kt
```diff
@@ -297,38 +296,134 @@ object ImageUtils {
...
+            sampleSize = 1
+        }
+
+        bmOptions.inSampleSize = sampleSize
+        bmOptions.inPurgeable = true
+        bmOptions.inPreferredConfig = Bitmap.Config.RGB_565
+        bmOptions.inJustDecodeBounds = false
+
+        /* Decode the JPEG file into a Bitmap */
+        val bitmap = BitmapFactory.decodeFile(path, bmOptions)
```
arunbakt, line 367: I am not very familiar with BitmapFactory. If the `sampleSize` is more than 1, will `BitmapFactory` downsize the image? For example if `imageWidth/requiredWidth` ratio is 2, will the bitmap created be half the width of the original?

app/src/main/java/org/greenstand/android/TreeTracker/utilities/ImageUtils.kt
```diff
@@ -297,38 +296,134 @@ object ImageUtils {
...
+
+        val rotatedBitmap = Bitmap.createBitmap(bitmap, 0, 0,
+            bmOptions.outWidth, bmOptions.outHeight, Matrix(), true)
+
+        val compressionQuality = 100
+        val byteArrayBitmapStream = ByteArrayOutputStream()
+        rotatedBitmap.compress(Bitmap.CompressFormat.PNG, compressionQuality,
```
arunbakt, line 374: What happens to the `byteArrayBitmapStream`, does it get flushed and closed by `rotatedBitmap.compress` method.  We don't seem to use it. I assume the verify first call to `BitmapFactory.decodeFile` writes the scaled version of the image and that is what we care about in this method. The rotatedBitmap is not used by the caller and neither the byteArrayBitmapStream.

Trying to reason with this code, not very familiar with it. Thank you.

arunbakt (commented):
Left some comments @Elforama. Please take a look and let me know your thoughts.

Elforama:
@arunbakt I'd like to address any issues in another PR. This is the old logic before the cameraX was used. I just moved everything to a new file.

app/src/main/java/org/greenstand/android/TreeTracker/activities/ImageCaptureActivity.kt
```diff
@@ -88,6 +84,7 @@ class ImageCaptureActivity : AppCompatActivity() {
                     }
 
                     override fun onImageSaved(file: File) {
+                        ImageUtils.resizedImage(file.absolutePath)
```
Elforama, line 87: @arunbakt This is the only line with any real changes. I resize the image after it's taken using existing resize logic.

app/src/main/java/org/greenstand/android/TreeTracker/utilities/ImageUtils.kt
Elforama: Not sure, but this is the old logic which is what we want.

app/src/main/java/org/greenstand/android/TreeTracker/utilities/ImageUtils.kt
Elforama: It's messy code. This is all legacy image scaling. I think we can clean this up in some later PRs. But for now we know that this code works.

app/src/main/java/org/greenstand/android/TreeTracker/activities/ImageCaptureActivity.kt
arunbakt: Got that, thank you!

Issue #463, opened 2020-08-15
Convergence parameter  for GPS data when adding a tree


arunbakt:
GPS location convergence parameter is the variation from the average that meets an allowed maximum threshold value below which a tree capture with device's camera is enabled.  When the planter intends to add a tree, tapping add a tree button should result in a spinner graphic while waiting for the convergence parameter. 

The convergence parameter is the greater of the latitude/longitude standard deviations of a set of recent GPS co-ordinates (ones captured from the time 'Add Tree' is clicked).  The set is last 'n' location ('n' is configurable) values  used in computing the convergence parameter (as standard deviation). When the convergence value computed with n set meets the max threshold, we let the Camera screen clickable to capture the tree. If not, we wait for the next n+1-n sequential  location value and compute the convergence parameter again. If this parameter is within the maximum allowed convergence threshold (configured as a property value), the planter is allowed to take the picture. If not, after a configurable timeout , the planter is allowed to take the picture anyways.

This issue is a subtask created from the requirements specified [here](https://github.com/Greenstand/Development-Overview/wiki/Android-v1.4-GPS-location-data-collection).

deepwinter:
accept capture when 

MAX(STDDEV(latitudes), STDDEV(longitudes)) < CONVERGENCE PARAMETER
where latitudes and longitudes are the last N latitudes and longitudes received from GPS respectively

arunbakt:
@deepwinter what if we have separate convergence threshold for latitude and longitude. Would that make the data closer to the true location of the tree.
i.e
  MAX(stddev(latitudes)) < latitude_convergence_parameter and
  MAX(stddev(longitude)) < longitude_convergence_parameter

arunbakt:
Issue addressed in the following PR and merged.
https://github.com/Greenstand/treetracker-android/pull/478

PR #478, opened 2020-09-03
Add tree with location convergence 

arunbakt:
This change addresses the requirement specified in https://github.com/Greenstand/treetracker-android/issues/463. 

Changes include:

- Compute standard deviation for longitude and latitude for a given location data size and determine if it is less than configured value (see `LocationDataConfig`) before allowing the planter to proceed with taking a picture of the tree

- If the convergence is not attained within the specified time, allow the planter to proceed with the picture anyways

- location data entity to record plain `json` string instead of base64 encoded values

Note: Please uninstall your existing treetracker app to install this version since there is no migration script path from previous commit in master.  But upgrade script from 1.3 to the current version is still valid.


app/src/main/java/org/greenstand/android/TreeTracker/utilities/LocationDataConfig.kt
```diff
@@ -0,0 +1,17 @@
+package org.greenstand.android.TreeTracker.utilities
+
+/**
+ *  app-wide constants and utilities
+ */
+object LocationDataConfig {
```
arunbakt, line 6: Eventually this would be refactored when UI configuration for controlling convergence parameter is built

app/src/main/java/org/greenstand/android/TreeTracker/fragments/NewTreeFragment.kt
```diff
@@ -13,25 +13,30 @@ import androidx.core.app.ActivityCompat
...
+import org.koin.android.ext.android.inject
 import org.koin.android.viewmodel.ext.android.viewModel
 import timber.log.Timber
 
 class NewTreeFragment :
     androidx.fragment.app.Fragment(), ActivityCompat.OnRequestPermissionsResultCallback {
 
     private val vm: NewTreeViewModel by viewModel()
+    private val locationDataCapturer by inject<LocationDataCapturer>()
+    private var newTreeUuid: UUID? = null
```
Elforama, line 39: I've reviewed this code already. Can you please merge the first PR and rebase this?

Elforama (changes requested):

Elforama:
Didn't realize until later that this seems to contain your work from this PR https://github.com/Greenstand/treetracker-android/pull/469/files

I ended up double reviewing a lot of code. Please rebase your work.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,329 @@
...
+import org.greenstand.android.TreeTracker.database.TreeTrackerDAO
+import org.greenstand.android.TreeTracker.database.entity.LocationDataEntity
+import org.greenstand.android.TreeTracker.utilities.LocationDataConfig
+import org.greenstand.android.TreeTracker.utilities.LocationDataConfig.CONVERGENCE_DATA_SIZE
+import org.greenstand.android.TreeTracker.utilities.LocationDataConfig.LAT_STD_DEV
+import org.greenstand.android.TreeTracker.utilities.LocationDataConfig.LONG_STD_DEV
+import org.greenstand.android.TreeTracker.utilities.ValueHelper
+import timber.log.Timber
+
+class LocationUpdateManager(
```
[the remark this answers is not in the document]
arunbakt: I have seen this model used in codebases where the language allows multiple classes in a file. The idea of grouping all related classes in a file appeals to me when they are not that big. Currently the number of lines is approx 300 lines.

One advantage of this is you  don't need to navigate to different files to get to the code that are related to each other. Optimized for reading. If you still feel separate files makes sense right now, let me know. Just wanted to let you know my reasoning.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,329 @@
...
+
+class LocationDataCapturer(
+    private val userManager: UserManager,
+    private val locationUpdateManager: LocationUpdateManager,
+    private val treeTrackerDAO: TreeTrackerDAO
+) {
+    var convergenceWithinRange: Boolean = false
+
+    private val gson = GsonBuilder().serializeNulls().create()
+    private var lastNLocations: Deque<Location> = LinkedList<Location>()
```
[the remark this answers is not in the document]
arunbakt: sure, I like the variable name indicating it as a queue as well in this case.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,329 @@
...
+    var generatedTreeUuid: UUID? = null
+        private set
+    var convergence: Convergence? = null
+
+    private val locationObserver: Observer<Location?> = Observer {
+        it?.apply {
```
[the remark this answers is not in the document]
arunbakt: Can I know why? In my IDE, I can see it is a Location.

app/src/main/java/org/greenstand/android/TreeTracker/utilities/LocationDataConfig.kt
```diff
@@ -0,0 +1,17 @@
...
+
+    // Location update config values
+    val MIN_TIME_BTWN_UPDATES = 1000L // Minimum time between location updates
+    val MIN_DISTANCE_BTW_UPDATES = 0F // Minimum distance between location updates
+
+    // Location Convergence config values
+    val CONVERGENCE_TIMEOUT = 60000L // Timeout value for Location Convergence
+    val CONVERGENCE_DATA_SIZE = 5 // Number of data required to compute convergence parameters
+    val LONG_STD_DEV = 0.00001 // Threshold standard deviation value for longitude convergence
+    val LAT_STD_DEV = 0.00001 // Threshold standard deviation value for latitude convergence
```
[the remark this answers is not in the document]
arunbakt: These will be revisited when we add a UI to control these values. It is a `val` inside a singleton object now, pretty close to constant.

app/src/main/java/org/greenstand/android/TreeTracker/utilities/BoundedConcurrentQueue.kt
```diff
@@ -0,0 +1,72 @@
+package org.greenstand.android.TreeTracker.utilities
+
+import java.util.Queue
+import java.util.concurrent.ConcurrentLinkedDeque
+
+class BoundedConcurrentQueue<E>(var capacity: Int) : Queue<E> {
```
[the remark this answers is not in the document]
arunbakt: This is what I added when I was overthinking this :).  Forgot to delete it.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,329 @@
...
+                isUpdating = true
+            }
+            true
+        } else {
+            isUpdating = false
+            false
+        }
+    }
+
+    fun stopLocationUpdates() {
```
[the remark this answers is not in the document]
arunbakt: Yeah, you are right. Will change it.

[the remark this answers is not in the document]
arunbakt: I am changing this to return a boolean to indicate if the request succeeds or not for client's who cares about it.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,329 @@
...
+import org.greenstand.android.TreeTracker.utilities.LocationDataConfig.LONG_STD_DEV
+import org.greenstand.android.TreeTracker.utilities.ValueHelper
+import timber.log.Timber
+
+class LocationUpdateManager(
+    private val locationManager: LocationManager,
+    private val context: Context
+) {
+
+    private val locationUpdates = MutableLiveData<Location?>()
```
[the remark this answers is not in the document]
arunbakt: This is the code that exists prior to my change. I suspect this is nullable because when the app starts it needs an initial value for the location which in this case is null.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,329 @@
...
+    var currentLocation: Location? = null
+        private set
+
+    init {
+        locationUpdates.postValue(null)
+    }
+
+    private val locationUpdateListener = object : android.location.LocationListener {
+
+        @UseExperimental(ExperimentalCoroutinesApi::class)
```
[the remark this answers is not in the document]
arunbakt: This is code from prior to my change. It is showing as added because I renamed the file.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,329 @@
+package org.greenstand.android.TreeTracker.managers
```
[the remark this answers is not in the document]
arunbakt: I brought all the location related classes inside `locations.kt` and thus LocationUpdateManager is removed.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/NewTreeFragment.kt
arunbakt: Ok, I will merge the first PR and rebase on this one.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,329 @@
...
+
+fun Location?.accuracyStatus(): Accuracy {
+    if (this == null || !hasAccuracy()) {
+        return Accuracy.NONE
+    }
+    return if (accuracy < ValueHelper.MIN_ACCURACY_DEFAULT_SETTING) {
+        Accuracy.GOOD
+    } else {
+        Accuracy.BAD
+    }
```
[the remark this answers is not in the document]
arunbakt: Yes, these are existing code. It shows as added because of restructuring the class `LocationUpdateManager.kt`

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,329 @@
...
+) {
+    var convergenceWithinRange: Boolean = false
+
+    private val gson = GsonBuilder().serializeNulls().create()
+    private var lastNLocations: Deque<Location> = LinkedList<Location>()
+    var generatedTreeUuid: UUID? = null
+        private set
+    var convergence: Convergence? = null
+
+    private val locationObserver: Observer<Location?> = Observer {
```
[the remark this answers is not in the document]
arunbakt: Here it is, https://developer.android.com/reference/androidx/lifecycle/Observer. You might have seen this already.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,329 @@
...
+                        generatedTreeUuid?.toString() ?: null,
+                        System.currentTimeMillis()
+                    )
+                Timber.d("Convergence: Generated Location Data value $locationData")
+                val jsonValue = gson.toJson(locationData)
+                Timber.d("Convergence: Inserting a new location data $jsonValue")
+                treeTrackerDAO.insertLocationData(LocationDataEntity(jsonValue))
+            }
+
+            if (isInTreeCaptureMode() && !(convergenceWithinRange)) {
```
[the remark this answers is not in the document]
arunbakt: Thank you, removing it

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,329 @@
...
+                var evictedLocation: Location? = if (lastNLocations.size >= CONVERGENCE_DATA_SIZE)
+                    lastNLocations.pollFirst() else null
+                lastNLocations.add(it)
+
+                if (lastNLocations.size >= CONVERGENCE_DATA_SIZE) {
+                    if (convergence == null) {
+                        convergence = Convergence(lastNLocations.toList())
+                        convergence!!.computeConvergence()
+                    } else {
+                        convergence!!.computeSlidingWindowConvergence(evictedLocation!!, it)
```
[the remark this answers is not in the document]
arunbakt: Evicted location won't be null  because `locationsDeque` will be of full size required for convergence for the initial computation. Subsequent computations would pop the head of the queue which becomes the evicted location as per line 173.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,329 @@
...
+
+class Convergence(val locations: List<Location>) {
+
+    var longitudeConvergence: ConvergenceStats? = null
+        private set
+    var latitudeConvergence: ConvergenceStats? = null
+        private set
+
+    fun computeStats(data: List<Double>): ConvergenceStats {
+        val mean = data.sum().div(data.size)
```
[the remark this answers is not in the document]
arunbakt: Sure, either one  is fine.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,329 @@
...
+     *  μn+1=μn−(1/99*X(n))+(1/99*X(n+100)) and
+     *  variance(n+1)= variance(n) −1/99(Xn−μn)^2 + 1/99(X(n+100)−μ(n+1))^2
+     */
+    fun computeSlidingWindowStats(
+        currentStats: ConvergenceStats,
+        replacingValue: Double,
+        newValue: Double
+    ): ConvergenceStats {
+        val newMean = currentStats.mean
+            .minus(replacingValue / CONVERGENCE_DATA_SIZE)
```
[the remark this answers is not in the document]
arunbakt: I thought it is neat to read, especially when the expression splits into multiple lines.  I don't have a strong preference either way.

arunbakt:
@Elforama Thank you for your feedback. I updated the PR with your recommended changes and left comments on others. Please take a look when you have a chance.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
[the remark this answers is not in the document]
Elforama: You can not use 'it' for any code in the rest of this block which can cause future bugs. The 'it' scope is very large.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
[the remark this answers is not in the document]
Elforama: thx

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,332 @@
...
+                var evictedLocation: Location? = if (locationsDeque.size >= CONVERGENCE_DATA_SIZE)
+                    locationsDeque.pollFirst() else null
+                locationsDeque.add(it)
+
+                if (locationsDeque.size >= CONVERGENCE_DATA_SIZE) {
+                    if (convergence == null) {
+                        convergence = Convergence(locationsDeque.toList())
+                        convergence!!.computeConvergence()
+                    } else {
+                        convergence!!.computeSlidingWindowConvergence(evictedLocation!!, it)
```
Elforama, line 184: I see 'it' is used here. It's hard to know what it is. This is also why I mentioned using a named variable at the top of this function.

Elforama:
Please rebase this before I review it again. You have 3 PRs (including mine) in this.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
arunbakt: Changed it use a named variable.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,329 @@
...
+            android.Manifest.permission.ACCESS_FINE_LOCATION
+        )
+        val coarseLocationPermission = ContextCompat.checkSelfPermission(
+            context,
+            android.Manifest.permission.ACCESS_COARSE_LOCATION
+        )
+        return (fineLocationPermission == PackageManager.PERMISSION_GRANTED ||
+                coarseLocationPermission == PackageManager.PERMISSION_GRANTED)
+    }
+}
```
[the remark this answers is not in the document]
arunbakt: These are existing code and not part of my change. It is being added because I restructured the Locations.kt to include the LocationUpdateManager

app/build.gradle
```diff
@@ -216,6 +213,7 @@ dependencies {
     implementation 'androidx.lifecycle:lifecycle-viewmodel-ktx:2.2.0-rc03'
     implementation 'androidx.lifecycle:lifecycle-runtime-ktx:2.2.0-rc03'
     implementation "androidx.work:work-runtime-ktx:2.2.0"
+    implementation "com.google.maps.android:android-maps-utils:2.0.3"
```
Elforama, line 216: What's this for?

app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
```diff
@@ -165,9 +169,18 @@ class MapsFragment : androidx.fragment.app.Fragment(), OnClickListener, OnMapRea
                             findNavController().navigate(
                                 MapsFragmentDirections.actionGlobalLoginFlowGraph())
                         } else {
+                            val progressBarLayout = view!!.findViewById<LinearLayout>(
```
Elforama, line 172: Please put view setup in onViewCreated or onCreateView

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,332 @@
...
+                        accuracy,
+                        generatedTreeUuid?.toString() ?: null,
+                        System.currentTimeMillis()
+                    )
+                val jsonValue = gson.toJson(locationData)
+                Timber.d("Inserting new location data $jsonValue")
+                treeTrackerDAO.insertLocationData(LocationDataEntity(jsonValue))
+            }
+
+            if (isInTreeCaptureMode() && !convergenceWithinRange) {
```
Elforama, line 173: I don't see where converenceWIthinRange is ever changed. is it?

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,332 @@
...
+
+                var evictedLocation: Location? = if (locationsDeque.size >= CONVERGENCE_DATA_SIZE)
+                    locationsDeque.pollFirst() else null
+                locationsDeque.add(it)
+
+                if (locationsDeque.size >= CONVERGENCE_DATA_SIZE) {
```
Elforama, line 179: This condition is already checked at line 175. What if this instead checks for evitedLocation nullablilty so it's safe to use it later on. That should remove it's !! on line 184

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
Elforama, line 184: Can we make this something like below to remove all the !!
`
convergence?.let {
   it.computeSlidingWIndow...
} ?: Convergence(...).computeConvergence()
`

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,332 @@
...
+                                "Longitude standard deviation value: " +
+                                "[${convergence!!.longitudeConvergence?.standardDeviation}]"
+                    )
+                    Timber.d(
+                        "Convergence: Latitude Mean: " +
+                                "[${convergence!!.latitudeConvergence?.mean}]. \n " +
+                                "Latitude standard deviation value: " +
+                                "[${convergence!!.latitudeConvergence?.standardDeviation}]"
+                    )
+                    safeLet(
```
Elforama, line 198: Can we remove this safeLet? It's only used in one place.

`
val lat = convergence?.longitudinalStandardDeviation() ?: return

val long = convergence?.latitudinalStandardDeviation() ?: return

convergenceWithinRange = long < LONG_STD_DEV && lat < LAT_STD_DEV
`

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,332 @@
...
+    var latitudeConvergence: ConvergenceStats? = null
+        private set
+
+    fun computeStats(data: List<Double>): ConvergenceStats {
+        val mean = data.sum() / data.size
+        var variance = 0.0
+        for (x in data) {
+            variance += (x - mean!!).pow(2.0)
+        }
+        variance /= data.size
```
Elforama, line 245: This is fine, but mutable. If we want to avoid that we can do:
`val variance = data.reduce { acc, value -> acc += (value - mean).pow(2.0) } / data.size`

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,332 @@
...
+        convergence = null
+        locationsDeque.clear()
+        convergenceWithinRange = false
+        Timber.d("Convergence: Tree capture turned off")
+    }
+}
+
+class Convergence(val locations: List<Location>) {
+
+    var longitudeConvergence: ConvergenceStats? = null
```
Elforama, line 234: I'm curious if it would help if there was an invalid ConvergenceStats state that this is initially set to instead of null. That way we can avoid all the double bands and null checks.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,332 @@
...
+data class LocationData(
+    val planterCheckInId: Long?,
+    val latitude: Double,
+    val longitude: Double,
+    val accuracy: Float,
+    val treeUuid: String?,
+    val capturedAt: Long
+)
+
+fun <T1 : Any, T2 : Any, R : Any> safeLet(p1: T1?, p2: T2?, block: (T1, T2) -> R?): R? {
```
Elforama, line 330: Let's see if we can remove this.

app/src/main/java/org/greenstand/android/TreeTracker/utilities/MathExtensions.kt
```diff
@@ -0,0 +1,4 @@
+package org.greenstand.android.TreeTracker.utilities
+
+fun Math.standardDeviation() {
```
Elforama, line 3: Can we remove this?

app/src/main/res/layout/fragment_map.xml
```diff
@@ -48,9 +48,32 @@
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintStart_toStartOf="parent"
         android:text="@string/gps_accuracy_double_colon"
-        android:textColor="@color/red"
+        android:textColor="@color/toolbar"
         android:textSize="@dimen/text_size_twenty" />
 
+    <LinearLayout
```
Elforama, line 54: The parent is a Constraint layout. Can we remove this linear layout from the view and use constraints instead?

app/src/test/java/org/greenstand/android/TreeTracker/managers/LocationDataCapturerTest.kt
```diff
@@ -56,14 +62,83 @@ class LocationDataCapturerTest {
...
+
+        assertFalse(locationDataCapturer.convergenceWithinRange)
+    }
+
+    val locationValuesLowVariance = listOf(
+        Pair(-122.08400001, 37.42149486),
+        Pair(-122.08400111, 37.42149487),
+        Pair(-122.08400021519, 37.42149483),
+        Pair(-122.08400021999, 37.42149489),
+        Pair(-122.0840086753, 37.42149481)
```
Elforama, line 101: This is fine, but it can also be done like:
`listOf(
123 to 456, 
678 to 901
)`

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
```diff
@@ -0,0 +1,332 @@
...
+     * recursive relation
+     *  μn+1=μn−(1/99*X(n))+(1/99*X(n+100)) and
+     *  variance(n+1)= variance(n) −1/99(Xn−μn)^2 + 1/99(X(n+100)−μ(n+1))^2
+     */
+    fun computeSlidingWindowStats(
+        currentStats: ConvergenceStats,
+        replacingValue: Double,
+        newValue: Double
+    ): ConvergenceStats {
+        val newMean = (
```
Elforama, line 269: Looks like there is an extra parenthesis for newMean and newVariance

Elforama (changes requested):

app/src/test/java/org/greenstand/android/TreeTracker/managers/LocationDataCapturerTest.kt
arunbakt: I don't understand what you are pointing to here. You want to use listOf instead of Pair?

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
arunbakt: The `if` condition checks the null value and it is fine to use `!!` in the else (gauranteed to be not null) and also is `terse`.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
arunbakt: It would not satisfy the primary intent that computation only occurs if the minimum size is reached. The very first computation would be missed if I base it solely on the nullability of `evictedLocation`.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
arunbakt: Yes, in line 202 it is flipped

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
arunbakt: I find this very useful and appropriate to do a null check on two variables. While this should be a utility function in a Util file,  it doesn't have multiple usages or other custom functions that we can group and move this out in its separate file yet.

Let me know if you have some other better way to replace it.

app/build.gradle
arunbakt: I was getting class not found error on a map related class (the name I couldn't recall now) when I was trying to run the app. Now that I tried without this, it works. Not sure exactly what happened. Removed this line now.

app/src/main/res/layout/fragment_map.xml
arunbakt: Using a constraint layout with Group accomplishes the same desired effect. Updating the code.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
Elforama: The nullability depends on the size so they should be the same. Shouldn't it?

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
Elforama: It's fine now, but in the future a NPE can be introduced. If there is a way to avoid !! I think it should be done.

Elforama: The reason why there is a !! is because it's not 100% guaranteed. it's possible the value is changed on another thread in the future. This is why Kotlin suggests using 'let' to check nullability.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
Elforama: I commented where this is used on a replacement.

This is something that we have to remember and each time it's seen it's another layer to understand and add complexity for other people.

Another way is to assign the two fields to local vals and do an if statement on them

val a = something?.thing()
val b = somethingelse?.thing()

if (a != null && b != null) {...}

app/src/main/res/layout/fragment_map.xml
```diff
@@ -48,7 +81,7 @@
         app:layout_constraintBottom_toBottomOf="parent"
         app:layout_constraintStart_toStartOf="parent"
         android:text="@string/gps_accuracy_double_colon"
-        android:textColor="@color/red"
+        android:textColor="@color/toolbar"
```
Elforama, line 84: The toolbar color should only be used for the toolbar. A new color can be used instead, if needed.

Elforama (changes requested):
Looking good overall. Just a few comments to address. 

It would be good to simplify some things and reduce the usage of '!!' as much as possible.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
arunbakt: Actually the very first time the size reaches (say 5) the required minimum, the `evictedLocation` is still null.  The line 177 will move the size of `locationsDeque` to 5 from 4 and this will be the very first computation and subsequent ones will evict the head and add the to the tail.

I have removed the `!!` and now using `?` within the if condition.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
arunbakt: I get that, updated  to use `convergence?` instead. Thank you.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
arunbakt: Sure, updating it with your suggestion.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
arunbakt: Updated this and removed `safeLet`

app/src/main/res/layout/fragment_map.xml
arunbakt: My bad, must have creeped when I was mucking with the design tool.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
arunbakt: This won't work since `acc += ` is not an expression and an assignment. Also, we need every value to be used to calculate the variance and this would miss the very first value as the call would use two values from the data missing the 0th index.

The var here is a local scope, so it is ok. If there is some other nice terse function to do this correctly, will update it.

app/src/main/java/org/greenstand/android/TreeTracker/managers/Locations.kt
arunbakt: Hmm, I see having null as appropriate. The only usage for these variables are behind a safe `?` operator. So we are good.

arunbakt (commented):
Updated the code to remove safeLet and use `?` avoiding `!!` wherever possible.

app/src/test/java/org/greenstand/android/TreeTracker/managers/LocationDataCapturerTest.kt
Elforama: I'm saying instead of
`Pair(a, b)`
you can do
`a to b`

Elforama (approved):
Looks good!

Elforama:
The variance code can be this if you want:
`val variance = list.fold(0.0) { acc, value -> acc + (value - mean).pow(2.0) } / list.size`

app/src/test/java/org/greenstand/android/TreeTracker/managers/LocationDataCapturerTest.kt
arunbakt: Just wanted a simple pair  of longitude, latitude values. Map would make the intent unclear.

arunbakt:
Thank you for the feedback @Elforama . Will merge this branch.

PR #479, opened 2020-09-04
Preferences cleanup

Elforama:
This replaces the old preferences usage with the new version.

Moved user based values into UserManager. 

Broke up Preferences file into many classes into their own package.

Replaced a lot of preference usage with UserManager

Migrates old key values to new keys. Deletes old key values


app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
```diff
@@ -55,17 +48,6 @@ class MainActivity : AppCompatActivity(), ActivityCompat.OnRequestPermissionsRes
 
         languageSwitcher.applyCurrentLanguage(this)
 
-        if (sharedPreferences.getBoolean(ValueHelper.FIRST_RUN, true)) {
```
arunbakt, line 58: I see this block of code has no consequences, thank you for deleting

app/src/main/java/org/greenstand/android/TreeTracker/usecases/ValidateCheckInStatusUseCase.kt
```diff
@@ -1,15 +1,14 @@
...
+import org.greenstand.android.TreeTracker.managers.UserManager
 
-class ValidateCheckInStatusUseCase constructor(private val sharedPreferences: SharedPreferences): UseCase<Unit, Boolean>() {
+class ValidateCheckInStatusUseCase constructor(private val userManager: UserManager) : UseCase<Unit, Boolean>() {
 
     override suspend fun execute(params: Unit): Boolean {
         val currentTimestampSeconds = System.currentTimeMillis() / 1000
 
-        val lastTimeStamp = sharedPreferences.getLong(ValueHelper.TIME_OF_LAST_PLANTER_CHECK_IN_SECONDS, 0)
+        val lastTimeStamp = userManager.lastCheckInTimeInSeconds ?: 0
```
arunbakt, line 11: I like this change promoting the model object. Reads better.

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/MapViewModel.kt
```diff
@@ -30,7 +28,7 @@ class MapViewModel constructor(
         if (validateCheckInStatusUseCase.execute(Unit)) {
             checkInStatusLiveData.postValue(true)
         } else {
-            expireCheckInStatusUseCase.execute(Unit)
+            userManager.clearUser()
```
arunbakt, line 31: Can we use some other intuitive name.  Instead of `clearUser`, we could `logOff` or `expireCheckInStatus`?

arunbakt (commented):
The change looks good to me overrall @Elforama, especially promoting `UserManager` in places where sharedPereferences were used directly. 

Left some comments and suggestions as well. Please take a look.

Thank you,
Arun

app/src/main/java/org/greenstand/android/TreeTracker/managers/UserManager.kt
```diff
@@ -1,47 +1,55 @@
...
+import org.greenstand.android.TreeTracker.preferences.PrefKey
+import org.greenstand.android.TreeTracker.preferences.PrefKeys
+import org.greenstand.android.TreeTracker.preferences.Preferences
+import org.koin.core.context.GlobalContext
 
-class UserManager(private val sharedPreferences: SharedPreferences) {
+class UserManager {
 
-    var authToken: String? = null
+    private val preferences: Preferences by GlobalContext.get().koin.inject()
```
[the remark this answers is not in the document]
Elforama: Unfortunately, UserManager and Preferences create a circular dependency. This injects it lazily to avoid the error. Def not clean...

app/src/main/java/org/greenstand/android/TreeTracker/managers/UserManager.kt
```diff
@@ -1,47 +1,55 @@
...
 
     companion object {
-        private const val FIRST_NAME_KEY = "FIRST_NAME_KEY"
-        private const val LAST_NAME_KEY = "LAST_NAME_KEY"
-        private const val ORG_NAME_KEY = "ORG_NAME_KEY"
+        private val BASE_KEY = PrefKeys.SESSION + PrefKey("info")
+        val PLANTER_CHECK_IN_ID_KEY = BASE_KEY + PrefKey("planter-check-in-id")
+        val PLANTER_INFO_ID_KEY = BASE_KEY + PrefKey("planter-info-id")
+        val ORG_NAME_KEY = BASE_KEY + PrefKey("organization")
+        val FIRST_NAME_KEY = BASE_KEY + PrefKey("first-name")
```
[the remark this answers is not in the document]
Elforama: UserPrefKey is for user unique values for that account. An example would be user A having language set to english and use B having it set to something else. PrefKey is for anything else not tied to a user.

A larger future use for UserPrefKeys will be the org flavor. A user will have specific screens enabled/disabled that will persist in shared prefs per each user.

In this case, these are session based values that are overridden per user on each login/logout.

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/MapViewModel.kt
Elforama: Good idea!

app/src/main/java/org/greenstand/android/TreeTracker/managers/UserManager.kt
```diff
@@ -1,47 +1,55 @@
...
 
-import android.content.SharedPreferences
-import org.greenstand.android.TreeTracker.utilities.ValueHelper
+import org.greenstand.android.TreeTracker.preferences.PrefKey
+import org.greenstand.android.TreeTracker.preferences.PrefKeys
+import org.greenstand.android.TreeTracker.preferences.Preferences
+import org.koin.core.context.GlobalContext
 
-class UserManager(private val sharedPreferences: SharedPreferences) {
+class UserManager {
```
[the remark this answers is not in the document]
Elforama: I like your thinking

app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
```diff
@@ -23,31 +23,21 @@ import com.google.android.gms.maps.OnMapReadyCallback
...
-import kotlinx.android.synthetic.main.activity_main.toolbarTitle
-import kotlinx.android.synthetic.main.fragment_map.addTreeButton
-import kotlinx.android.synthetic.main.fragment_map.goToUploadsButton
-import kotlinx.android.synthetic.main.fragment_map.mapUserImage
-import kotlinx.coroutines.Dispatchers
-import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.launch
-import kotlinx.coroutines.runBlocking
-import kotlinx.coroutines.withContext
+import kotlinx.android.synthetic.main.activity_main.*
```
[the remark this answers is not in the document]
Elforama: I fixed the other, but would like to address this in another PR. I'd like to move away from the synthetic views in our app.

Elforama:
I fixed some wildcard imports but it looks like my IDE changes it back. I think once I fix my IDE I will make a new PR to fix wildcard imports.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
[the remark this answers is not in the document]
arunbakt: The only issue is that, if I happen to touch this file, I would be forced to add the classes individually.

app/src/main/java/org/greenstand/android/TreeTracker/managers/UserManager.kt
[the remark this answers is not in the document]
arunbakt: The only usage for `UserManager` inside `Preferences` is accessing `platnerInfoId`. Could we create a `UserIdentity` data class with `planterInfoId` and pass it to `Preferences`. 

Another possibility is make computePath implemented in `PrefKey` and overriden in `UserPrefKey`. Only `UserPrefKey` will depend on `UserManager` or `UserIdentity`. In this case `Preferences` doesn't need to know `UserManager`.

app/src/main/java/org/greenstand/android/TreeTracker/managers/UserManager.kt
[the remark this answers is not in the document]
arunbakt: So, UserPrefKeys are persisting the preferences across sessions. Even when UserA logs out, his preferences stored with preferences will be retained but his sessions keys will be reset. Thank you. It will be nice to have some comments on UserPrefKey and PrefKey classes for onboarding developers.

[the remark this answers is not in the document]
Elforama: I'll add some docs to the code and wiki later.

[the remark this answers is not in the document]
Elforama: Sorry, I misspoke. UserPrefKeys are for sessions and anything else. They are simply tied to a specific user. In this case the values like first_name don't need to be tied to a user. They are reset on each logout/login. No more than 1 of those values exists at a time.

app/src/main/java/org/greenstand/android/TreeTracker/managers/LanguageSwitcher.kt
```diff
@@ -26,7 +29,7 @@ enum class Language(val locale: Locale) {
 class LanguageSwitcher(private val prefs: Preferences) {
 
     fun applyCurrentLanguage(activity: Activity) {
-        val language = Language.fromString(prefs.getString(LANGUAGE_PREF_KEY, "en"))
+        val language = Language.fromString(prefs.getString(LANGUAGE_PREF_KEY, "en") ?: "")
```
arunbakt, line 32: Why do we need `?: ""` here?  I see the default value of "en" should  take hold if `getString` has null value set. Am I missing to see the need?

app/src/main/java/org/greenstand/android/TreeTracker/preferences/Preferences.kt
```diff
@@ -1,71 +1,65 @@
...
+    fun getInt(prefKey: PrefKey, default: Int = -1): Int {
         return prefs.getInt(computePath(prefKey), default)
     }
 
-    fun computePath(prefKey: PrefKey): String {
+    private fun computePath(prefKey: PrefKey): String {
         return when (prefKey) {
             is UserPrefKey -> {
-                prefKey.path + "/${userManager.planterInfoId}"
+                prefKey.path + "/${_planterInfoId}"
```
arunbakt, line 39: It would be intuitive to have `_planterInfoId` within `UserPrefKey`, given it is associated with a specific `User`.

app/src/main/java/org/greenstand/android/TreeTracker/managers/User.kt
```diff
@@ -0,0 +1,56 @@
...
+    var lastCheckInTimeInSeconds: Long?
+        get() = preferences.getLong(LAST_CHECK_IN_TIME_IN_KEY)
+        set(value) = preferences.edit().putLong(LAST_CHECK_IN_TIME_IN_KEY, value ?: -1).apply()
+
+    fun expireCheckInStatus() = preferences.clearPrefKeyUsage(BASE_KEY)
+
+    companion object {
+        private val BASE_KEY = PrefKeys.SESSION + PrefKey("info")
+        val PLANTER_CHECK_IN_ID_KEY = BASE_KEY + PrefKey("planter-check-in-id")
+        val PLANTER_INFO_ID_KEY = BASE_KEY + PrefKey("planter-info-id")
```
arunbakt, line 49: It would be a bit more intuitive if we have `PLANTER_INFO_ID_KEY` derived from `UserPrefKey` and to omit line 38. This in combination with line 39 in `computePath` to use `UserPrefKey` will make it that much more cohesive.

arunbakt (commented):
@Elforama , Looks good to me. Left couple of comments to see what you think about using UserPrefKey.

app/src/main/java/org/greenstand/android/TreeTracker/preferences/Preferences.kt
Elforama: If we do this, then we'd need to recreate all the keys every time someone logs out. Otherwise the new user would have UserPrefKeys with the previous users ID associated with them.

app/src/main/java/org/greenstand/android/TreeTracker/managers/User.kt
Elforama: The User class holds the current user data **for the app**. So these preferences are not associated with an individual user.

app/src/main/java/org/greenstand/android/TreeTracker/managers/LanguageSwitcher.kt
Elforama: The default value can be null, so I need to cover that case.

Issue #464, opened 2020-08-17
Capture step counts when saving tree image


arunbakt:
Capture the step counts when saving a tree image in the 'Add tree' flow. Store both the absolute value of the step count returned by sensor and relative value (difference between the absolute value of the step count captured when last tree was saved and the tree currently being saved).

The upload process to send tree data should include the newly added step counter attributes as part of this implementation.

This issue is part of the requirement specified [here](https://github.com/Greenstand/treetracker-android/wiki/Device-Data-Capture)

PR #487, opened 2020-09-11
Capture step counts on tree capture

arunbakt:
This change addresses the Issue [464](https://github.com/Greenstand/treetracker-android/issues/464).

Changes include:
- Register step counter sensor in MainActivity and store the absolute steps for every change event in user object. 
- Capture the current absolute steps when capturing a tree. Find the delta using the step count from previous tree capture.
- Store the absolute step count and delta step count in the tree attributes when saving tree capture
- Refactor `TreeManager` to `Tree` and rename `manager` package to `model`
- Parcelize the `Tree` model and remove`NewTree` data since it is redundant


app/src/main/AndroidManifest.xml
```diff
@@ -7,17 +7,13 @@
         android:glEsVersion="0x00020000"
         android:required="true"/>
 
-    <uses-feature
```
arunbakt, line 10: I see the following error in my IDE on this line.  I removed this entry since the effect of marking the attribute `required` as false is equivalent to not stating the hardware use. Let me know what you think.

`Expecting android:required="false" for this hardware feature that may not be supported by all Chrome OS devices`.

app/src/main/AndroidManifest.xml
```diff
@@ -26,11 +22,7 @@
     <uses-permission android:name="android.permission.ACCESS_COARSE_LOCATION"/>
     <uses-permission android:name="android.permission.ACCESS_FINE_LOCATION"/>
     <uses-permission android:name="android.permission.ACCESS_NETWORK_STATE"/>
-
-    <permission
```
arunbakt, line 30: Saw this entry in stackoverflow. Hence removed it. https://stackoverflow.com/questions/14832911/android-map-v2-why-maps-receive-permission

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
```diff
@@ -232,4 +251,15 @@ class MainActivity : AppCompatActivity(), ActivityCompat.OnRequestPermissionsRes
         locationUpdateManager.startLocationUpdates()
         locationDataCapturer.start()
     }
+
+    override fun onAccuracyChanged(sensor: Sensor?, accuracy: Int) {
+        // Ignore
```
arunbakt, line 256: Wish there was a default implementation for this making it unnecessary to declare it here. Not sure if we  really care about change in accuracy reported for step count. Ignoring the implementation.

app/src/main/java/org/greenstand/android/TreeTracker/models/User.kt
```diff
@@ -39,7 +38,17 @@ class User(
 
     var lastCheckInTimeInSeconds: Long?
         get() = preferences.getLong(LAST_CHECK_IN_TIME_IN_KEY)
-        set(value) = preferences.edit().putLong(LAST_CHECK_IN_TIME_IN_KEY, value ?: -1).apply()
+        set(value) = preferences
+            .edit().putLong(LAST_CHECK_IN_TIME_IN_KEY, value ?: -1).apply()
+
+    var absoluteStepCount: Int?
```
arunbakt, line 44: I am assuming the collected step count is relevant for a specific user session when the planter is capturing trees. Hence adding these in the User model so that they all are removed when user switches or session ends.

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/NewTreeViewModel.kt
```diff
@@ -52,7 +51,21 @@ class NewTreeViewModel(
...
+            content = note,
+            photoPath = photoPath!!
+        )
+
+        val absoluteStepCount = user.absoluteStepCount ?: 0
+        val lastStepCountWhenCreatingTree = user.absoluteStepCountOnTreeCapture ?: 0
+        // Delta step count is the difference between the absolute count at the time of capturing
+        // a tree minus the last absolute step count recorded when capturing a previous tree. This
+        // is the indicator for the number of steps taken between two trees.
+        val deltaSteps = absoluteStepCount - lastStepCountWhenCreatingTree
```
Elforama, line 66: Looking at the docs it seems the step counter shows total steps since we started tracking since the device has last turned on. That sounds like if we plant a tree, then wait a week we can have a delta of many miles for our next tree. Is this correct?

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
```diff
@@ -232,4 +251,15 @@ class MainActivity : AppCompatActivity(), ActivityCompat.OnRequestPermissionsRes
...
+    }
+
+    override fun onSensorChanged(event: SensorEvent?) {
```
Elforama, line 259: The [docs](https://developer.android.com/guide/topics/sensors/sensors_motion) suggest using the job scheduler to query the current step count.

Elforama (changes requested):
Hi Arun, I've left some comments. Please let me know if you have any questions. 

Also, I'd really like if all the refactoring can be moved. If it's too much, then we can skip that for now, but in the future please leave formatting/refactoring to their own PRs. If you decide to skip it for this, please comment on files such as the treeentity which lines you changed.

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
arunbakt: I did see that. But I saw this example in the documentation for `SensorManager` in android developer guide. I wish there was an example for using JobScheduler.

app/src/main/java/org/greenstand/android/TreeTracker/models/User.kt
arunbakt: Yes, that can be done. But isn't `Session` inherently `User` associated?

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
```diff
@@ -18,16 +23,23 @@ import androidx.core.content.ContextCompat
...
+import org.greenstand.android.TreeTracker.models.LanguageSwitcher
+import org.greenstand.android.TreeTracker.models.LocationDataCapturer
+import org.greenstand.android.TreeTracker.models.LocationUpdateManager
+import org.greenstand.android.TreeTracker.models.User
 import org.koin.android.ext.android.inject
+import timber.log.Timber
 
-class MainActivity : AppCompatActivity(), ActivityCompat.OnRequestPermissionsResultCallback {
+class MainActivity : AppCompatActivity(), ActivityCompat.OnRequestPermissionsResultCallback,
+    SensorEventListener {
```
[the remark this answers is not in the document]
arunbakt: Sure, will do that.

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
```diff
@@ -57,7 +73,8 @@ class MainActivity : AppCompatActivity(), ActivityCompat.OnRequestPermissionsRes
 
         val listener = NavController
             .OnDestinationChangedListener { controller, destination, arguments ->
-                if (destination.id != R.id.splashFragment2 && destination.id != R.id.orgWallFragment) {
+                if (destination.id != R.id.splashFragment2 &&
```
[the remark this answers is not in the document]
arunbakt: @Elforama , This is one of those PRs where restructuring has to go hand in hand with the change. I could have left out the step count specific changes but it wouldn't have made the need apparent. Actually the code for step count change is minimal.

The formatting is done automatically on commit by ktlint. These non-conformant formatting are from our legacy code base and as we progress (with everyone having ktlint on) we should see these go down drastically overtime.

I will make sure I have enough comments on the PR pointing to the central change. 

Also, I tested the app with data verification to give some assurance.

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
```diff
@@ -127,6 +144,7 @@ class MainActivity : AppCompatActivity(), ActivityCompat.OnRequestPermissionsRes
     public override fun onPause() {
         super.onPause()
         locationUpdateManager.stopLocationUpdates()
+        sensorManager.unregisterListener(this)
```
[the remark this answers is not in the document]
arunbakt: I don't understand what you are referring to here. Can you elaborate please?

app/src/main/java/org/greenstand/android/TreeTracker/database/TreeTrackerDAO.kt
```diff
@@ -99,10 +99,6 @@ interface TreeTrackerDAO {
     @Query("SELECT * FROM tree_capture WHERE _id IN (:ids)")
     fun getTreeCapturesByIds(ids: List<Long>): List<TreeCaptureEntity>
 
-//    @Transaction
-//    @Query("SELECT * FROM tree_capture WHERE _id = :id")
-//    fun getTreeUploadDataById(id: Long): TreeUploadDbView
-
```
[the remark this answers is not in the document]
arunbakt: I usually removed commented codes I come across.

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
Elforama: I think it would be worth it to find out away to do this with the JobScheduler since it mentions it can drain the battery otherwise.

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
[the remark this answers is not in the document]
Elforama: https://developer.android.com/topic/libraries/architecture/lifecycle

app/src/main/java/org/greenstand/android/TreeTracker/models/User.kt
Elforama: They are associated, but they don't have to be in the same class. When I think of checking my current step count, i'd look for a step count class, not the user class.

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
arunbakt: I thought there would be an API like you are suggesting in the beginning. But there isn't one.  Actually, the only way to get a step count value is to listening to SensorEvent. At least to my knowledge. 

Thus, our use case needs this value at regular intervals. Of the three cadences available, `SENSOR_DELAY_NORMAL` seems more friendly to battery usage while maintaining a level of accuracy. The other choices are `SENSOR_DELAY_FASTEST`, `SENSOR_DELAY_GAME`, `SENSOR_DELAY_UI`. 

 I believe the current implementation won't be any more draining than a JobScheduler implementation since we are going to get frequent intervals of reading the value from the change event anyways.

Elforama: What I mean by the API suggestion is to make one like that ourselves to wrap what android supplies. We don't need constant updates of the step count, we only need the step count at the time that we plant the tree, correct?

So instead of getting a continuous stream of step updates, we can listen to step events until we get one step update, then we unsubscribe and use that value for the calculation.

The current implementation will get step updates even if the user is just looking at the map or syncing data.

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/NewTreeViewModel.kt
arunbakt: That is true. This came up in our dicussion in the team meeting and @deepwinter mentioned it is fine for the very first tree captured in a session. They can factor that in.

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
arunbakt: I made a change now that uses API to enable the step count listener and collect the values at the time of interest and disable the listener. In addition created an observer on destroy lifecyle event to unregister the listener when the app is exited without completing the cycle enabling/disabling the listener.

app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
```diff
@@ -139,6 +146,11 @@ class MainActivity : AppCompatActivity(), ActivityCompat.OnRequestPermissionsRes
         }
     }
 
+    public override fun onDestroy() {
```
arunbakt, line 149: This is added to ensure `stepCounter.viewLifecycleObserver` is removed when the app is exited unexpectedly without completing the tree capture.

app/src/main/java/org/greenstand/android/TreeTracker/activities/SplashFragment.kt
```diff
@@ -13,8 +13,8 @@ import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import org.greenstand.android.TreeTracker.BuildConfig
 import org.greenstand.android.TreeTracker.R
-import org.greenstand.android.TreeTracker.managers.FeatureFlags
-import org.greenstand.android.TreeTracker.managers.User
+import org.greenstand.android.TreeTracker.models.FeatureFlags
+import org.greenstand.android.TreeTracker.models.User
```
arunbakt, line 17: Please ignore, these are changes due to package rename. Not the main change for this PR.

app/src/main/java/org/greenstand/android/TreeTracker/analytics/Analytics.kt
```diff
@@ -16,7 +16,7 @@ import org.greenstand.android.TreeTracker.analytics.AnalyticEvents.USER_ENTERED_
 import org.greenstand.android.TreeTracker.analytics.AnalyticEvents.USER_ENTERED_EMAIL_PHONE
 import org.greenstand.android.TreeTracker.analytics.AnalyticEvents.USER_INFO_CREATED
 import org.greenstand.android.TreeTracker.data.TreeColor
-import org.greenstand.android.TreeTracker.managers.User
+import org.greenstand.android.TreeTracker.models.User
```
arunbakt, line 19: Please ignore

app/src/main/java/org/greenstand/android/TreeTracker/application/TreeTrackerApplication.kt
```diff
@@ -40,8 +40,10 @@ class TreeTrackerApplication : Application() {
         if (BuildConfig.DEBUG) {
             Timber.plant(Timber.DebugTree())
 
-            Timber.tag("DebugDB").d("To forward DebugDB from emulator to browser use the command 'adb forward tcp:8080 tcp:8080' from terminal")
-            Timber.tag("DebugDB").d("For more information visit: https://github.com/amitshekhariitbhu/Android-Debug-Database")
+            Timber.tag("DebugDB").d("To forward DebugDB from emulator to browser " +
```
arunbakt, line 43: Formatting changes, ignore

app/src/main/java/org/greenstand/android/TreeTracker/data/TreeHeightAttributes.kt
```diff
@@ -5,8 +5,9 @@ import kotlinx.android.parcel.Parcelize
 import org.greenstand.android.TreeTracker.BuildConfig
 
 @Parcelize
-data class TreeHeightAttributes(val treeId: Long,
-                                val heightColor: TreeColor,
-                                val appFlavor: String = BuildConfig.FLAVOR,
-                                val appBuild: String = BuildConfig.VERSION_NAME,
-                                val appVersion: String = BuildConfig.VERSION_CODE.toString()): Parcelable
\ No newline at end of file
+data class TreeHeightAttributes(
```
arunbakt, line 8: Formatting changes, ignore

app/src/main/java/org/greenstand/android/TreeTracker/database/TreeTrackerDAO.kt
```diff
@@ -136,6 +132,19 @@ interface TreeTrackerDAO {
...
+        attributes: List<TreeAttributeEntity>?
+    ): Long {
+        val treeId = insertTreeCapture(tree)
+        attributes?.forEach {
+            it.treeCaptureId = treeId
+            insertTreeAttribute(it)
+        }
+        return treeId
+    }
+
```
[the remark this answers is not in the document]
arunbakt: This is needed because we want both the inserts succeed together in a transaction. @Transaction I suspect is not supported on any non `Dao` interface or classes. Please correct me if I am wrong.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
```diff
@@ -37,12 +37,12 @@ import kotlinx.coroutines.withContext
 import kotlinx.coroutines.withTimeoutOrNull
 import org.greenstand.android.TreeTracker.R
 import org.greenstand.android.TreeTracker.database.TreeTrackerDAO
-import org.greenstand.android.TreeTracker.managers.Accuracy
-import org.greenstand.android.TreeTracker.managers.FeatureFlags
-import org.greenstand.android.TreeTracker.managers.LocationUpdateManager
-import org.greenstand.android.TreeTracker.managers.User
-import org.greenstand.android.TreeTracker.managers.accuracyStatus
 import org.greenstand.android.TreeTracker.map.TreeMapMarker
+import org.greenstand.android.TreeTracker.models.Accuracy
```
arunbakt, line 41: Ignore

app/src/main/java/org/greenstand/android/TreeTracker/fragments/NewTreeFragment.kt
```diff
@@ -18,7 +18,7 @@ import kotlinx.android.synthetic.main.fragment_new_tree.*
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.launch
 import org.greenstand.android.TreeTracker.R
-import org.greenstand.android.TreeTracker.managers.FeatureFlags
+import org.greenstand.android.TreeTracker.models.FeatureFlags
```
arunbakt, line 21: Ignore

app/src/main/java/org/greenstand/android/TreeTracker/models/FeatureFlags.kt
```diff
@@ -1,4 +1,4 @@
-package org.greenstand.android.TreeTracker.managers
+package org.greenstand.android.TreeTracker.models
```
arunbakt, line 1: Ignore

app/src/main/java/org/greenstand/android/TreeTracker/models/StepCounter.kt
```diff
@@ -0,0 +1,63 @@
+package org.greenstand.android.TreeTracker.models
```
arunbakt, line 1: This is the class that incorporates your suggestion to isolate the step count collection in its own class.

app/src/main/java/org/greenstand/android/TreeTracker/models/User.kt
```diff
@@ -39,7 +38,8 @@ class User(
 
     var lastCheckInTimeInSeconds: Long?
         get() = preferences.getLong(LAST_CHECK_IN_TIME_IN_KEY)
-        set(value) = preferences.edit().putLong(LAST_CHECK_IN_TIME_IN_KEY, value ?: -1).apply()
+        set(value) = preferences
```
arunbakt, line 41: Ignore formatting changes

app/src/main/java/org/greenstand/android/TreeTracker/usecases/ValidateCheckInStatusUseCase.kt
```diff
@@ -12,7 +12,8 @@ class ValidateCheckInStatusUseCase constructor(private val user: User) : UseCase
 
         val timeSinceLastCheckIn = currentTimestampSeconds - lastTimeStamp
 
-        if (FeatureFlags.AUTOMATIC_SIGN_OUT_FEATURE_ENABLED && timeSinceLastCheckIn > CHECK_IN_TIMEOUT) {
+        if (FeatureFlags.AUTOMATIC_SIGN_OUT_FEATURE_ENABLED &&
```
arunbakt, line 15: Ignore, just formatting changes here

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/MapViewModel.kt
```diff
@@ -50,6 +52,7 @@ class MapViewModel constructor(
 
     suspend fun turnOnTreeCaptureMode() {
         locationDataCapturer.turnOnTreeCaptureMode()
+        stepCounter.enable()
```
arunbakt, line 55: Step counter listener registration happens at this point

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/NewTreeViewModel.kt
```diff
@@ -61,23 +76,16 @@ class NewTreeViewModel(
         if (FeatureFlags.TREE_HEIGHT_FEATURE_ENABLED) {
             navigateToTreeHeight.postValue(newTree)
         } else {
-            saveTree(newTree)
+            createTreeUseCase.execute(newTree)
+            // Assign the current absolute step count to 'absoluteStepCountOnTreeCapture' to
+            // enable step count delta calculation for the next tree capture
+            stepCounter.absoluteStepCountOnTreeCapture = absoluteStepCount
+            stepCounter.disable()
```
arunbakt, line 83: StepCount data has been collected by now. Disaling causes the listener to be unregistered.

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/NewTreeViewModel.kt
```diff
@@ -61,23 +76,16 @@ class NewTreeViewModel(
...
             onTreeSaved.postValue(Unit)
             navigateBack.postValue(Unit)
         }
     }
 
-    private suspend fun saveTree(newTree: NewTree): Long {
```
arunbakt, line 70: Unused methods, `Tree` in models package takes over the functionality that was supported by `NewTree`

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/TreeHeightViewModel.kt
```diff
@@ -45,38 +44,27 @@ class TreeHeightViewModel(
...
                     }
                 }
                 ?.also {
                     toastMessageLiveData.postValue(R.string.tree_saved)
                     analytics.treeHeightMeasured(treeColor!!)
                     onFinishedLiveData.postValue(Unit)
+                    // Assign the current absolute step count to 'absoluteStepCountOnTreeCapture'
+                    // variable to calculate future step count deltas
+                    stepCounter.absoluteStepCountOnTreeCapture = stepCounter.absoluteStepCount
+                    stepCounter.disable()
```
arunbakt, line 67: Same operations to collect data and unregister the step count event listener is made for the use case where Tree Height attribute feature is enabled in the App

arunbakt (commented):
@Elforama Incorporated your suggestions and left pointers where the changes are. Let me know what you think when you get a chance. Thank you!

app/src/main/java/org/greenstand/android/TreeTracker/models/StepCounter.kt
```diff
@@ -0,0 +1,63 @@
...
+        get() = preferences.getInt(User.ABS_STEP_COUNT_ON_TREE_CAPTURE)
+        set(value) = preferences
+            .edit().putInt(User.ABS_STEP_COUNT_ON_TREE_CAPTURE, value ?: 0).apply()
+
+    val viewLifecycleObserver: LifecycleObserver = object : LifecycleObserver {
+
+        @OnLifecycleEvent(Lifecycle.Event.ON_DESTROY)
+        fun unregisterListener() {
+            Timber.d("StepCounter: unregister listener")
+            sensorManager.unregisterListener(stepCountEventListener)
```
Elforama, line 36: We can call disable here instead

app/src/main/java/org/greenstand/android/TreeTracker/models/StepCounter.kt
```diff
@@ -0,0 +1,63 @@
...
+    private var enabled = false
+
+    var absoluteStepCount: Int?
+        get() = preferences.getInt(User.ABS_STEP_COUNT)
+        private set(value) = preferences.edit().putInt(User.ABS_STEP_COUNT, value ?: 0).apply()
+
+    var absoluteStepCountOnTreeCapture: Int?
+        get() = preferences.getInt(User.ABS_STEP_COUNT_ON_TREE_CAPTURE)
+        set(value) = preferences
+            .edit().putInt(User.ABS_STEP_COUNT_ON_TREE_CAPTURE, value ?: 0).apply()
```
Elforama, line 29: Can we move these PrefKeys into this class?

app/src/main/java/org/greenstand/android/TreeTracker/models/StepCounter.kt
```diff
@@ -0,0 +1,63 @@
...
+import timber.log.Timber
+
+class StepCounter(
+    private val sensorManager: SensorManager,
+    private val preferences: Preferences
+) {
+
+    private val stepCounter = sensorManager.getDefaultSensor(Sensor.TYPE_STEP_COUNTER)
+    private val stepCountEventListener = StepCountEventListener()
+    private var enabled = false
```
Elforama, line 20: I don't see this used anywhere.

app/src/main/java/org/greenstand/android/TreeTracker/models/StepCounter.kt
Elforama: Or we could make the class implement LifecycleObserver and just annotate the 'disable' method

Elforama (changes requested):

Elforama:
Looks great, just some minor comments.

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/NewTreeViewModel.kt
deepwinter: Yes this is fine.  We will catch that in the pipeline.

PR #492, opened 2020-09-17
Preferences session cleanup

Elforama:
Cleanup naming of PrefKeys.
Update code so that everything depending on SESSION prefkey is deleted once the user logs out.


app/src/main/java/org/greenstand/android/TreeTracker/preferences/Preferences.kt
```diff
@@ -35,16 +35,16 @@ class Preferences(
 
     private fun computePath(prefKey: PrefKey): String {
         return when (prefKey) {
-            is UserPrefKey -> {
-                prefKey.path + "/${_planterInfoId}"
-            }
-            else -> {
-                prefKey.path
-            }
+            is UserPrefKey -> prefKey.path + "/${_planterInfoId}"
```
arunbakt, line 38: It will be nice to have a documentation with e.g. cases where `_planterInfoId` could be used in the keypath. For instance, `UserPrefKey` is used when setting language preference. This is a scenario where the preference is persisted across Sessions correct? So If you could add documentation for `UserPrefKey` that would throw some light when someonelooks at `PrefKey` and `UserPrefKey` and wonder when to use them.

arunbakt (approved):
Looks good to me @Elforama . Left a comment stating where additional documentation would be useful.

Arun

app/src/main/java/org/greenstand/android/TreeTracker/preferences/Preferences.kt
Elforama: Sorry, forgot to link to the actual docs I wrote. I can add a link later into the code to this wiki page: https://github.com/Greenstand/treetracker-android/wiki/Preferences-&-PrefKeys

PR #493, opened 2020-09-17
Add convergence indicator to location data

arunbakt:
The change is to add a field in location data to indicate whether the location collected during tree capture is converged or if the request for convergence is timed out.  Since location data is also collected outside the scope of tree capture, the convergence status in those cases will be null.

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
@@ -30,12 +30,9 @@ import kotlinx.android.synthetic.main.fragment_map.goToUploadsButton
 import kotlinx.android.synthetic.main.fragment_map.mapUserImage
 import kotlinx.coroutines.Dispatchers
 import kotlinx.coroutines.GlobalScope
-import kotlinx.coroutines.TimeoutCancellationException
-import kotlinx.coroutines.delay
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.withContext
-import kotlinx.coroutines.withTimeout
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/fragments/MapsFragment.kt
```diff
@@ -171,10 +172,15 @@ class MapsFragment : androidx.fragment.app.Fragment(), OnClickListener, OnMapRea
                         } else {
                             convergenceProgressView.visibility = View.VISIBLE
                             vm.turnOnTreeCaptureMode()
-                            withTimeoutOrNull(LocationDataConfig.CONVERGENCE_TIMEOUT) {
-                                while (!vm.isConvergenceWithinRange()) {
-                                    delay(LocationDataConfig.MIN_TIME_BTWN_UPDATES)
+                            try {
```
Elforama, line 175: We can move all this to the viewmodel

Elforama (changes requested):
Looks good, just one comment.

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/MapViewModel.kt
```diff
@@ -50,10 +55,21 @@ class MapViewModel constructor(
...
     }
 
-    suspend fun turnOnTreeCaptureMode() {
+    fun turnOnTreeCaptureMode() {
         locationDataCapturer.turnOnTreeCaptureMode()
         stepCounter.enable()
     }
 
-    fun isConvergenceWithinRange() = locationDataCapturer.convergenceWithinRange
+    suspend fun resolveLocationConvergence() {
```
arunbakt, line 63: Moving the timeout logic into the view model as you suggested

Issue #465, opened 2020-08-17
Capture compass heading and device orientation when tree image is captured


arunbakt:
Capture compass heading and device orientation and store the values in the tree attributes entity when a tree image is captured. Update the upload process to include the heading and orientation data.

The reasoning and requirement for this  is detailed [here](https://github.com/Greenstand/treetracker-android/wiki/Device-Data-Capture)

PR #501, opened 2020-09-23
Capture device orientation by using rotation vector sensor

arunbakt:
The change adds the feature to capture device orientation data (Rotation matrix) using Rotation Vector Sensor.
This addresses the issue https://github.com/Greenstand/treetracker-android/issues/465.

Change includes:
- Capturing the snapshot of the rotation matrix when the camera button is clicked
- The rotation matrix is a float array of length size 16
- The captured rotation matrix snapshot is added a tree attribute as a string with values seperated by commas

The change is tested in a samsung galaxy s8 device

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/activities/ImageCaptureActivity.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/activities/ImageCaptureActivity.kt
@@ -9,20 +9,27 @@ import android.view.TextureView
 import android.widget.ImageButton
 import android.widget.TextView
 import androidx.appcompat.app.AppCompatActivity
-import androidx.camera.core.*
+import androidx.camera.core.CameraX
+import androidx.camera.core.ImageCapture
+import androidx.camera.core.ImageCaptureConfig
+import androidx.camera.core.Preview
+import androidx.camera.core.PreviewConfig
+import java.io.File
 import org.greenstand.android.TreeTracker.R
+import org.greenstand.android.TreeTracker.models.DeviceOrientation
 import org.greenstand.android.TreeTracker.utilities.AutoFitPreviewBuilder
 import org.greenstand.android.TreeTracker.utilities.ImageUtils
 import org.greenstand.android.TreeTracker.utilities.ValueHelper
 import org.greenstand.android.TreeTracker.viewmodels.NewTreeViewModel.Companion.FOCUS_THRESHOLD
+import org.koin.android.ext.android.inject
 import timber.log.Timber
-import java.io.File
 
 class ImageCaptureActivity : AppCompatActivity() {
 
     private lateinit var viewFinder: TextureView
     private lateinit var imageCaptureButton: ImageButton
     private lateinit var toolbarTitle: TextView
+    private val deviceOrientation by inject<DeviceOrientation>()
 
     companion object {
         private const val SELFIE_MODE = "SELFIE_MODE"
@@ -71,6 +78,7 @@ class ImageCaptureActivity : AppCompatActivity() {
         // Build the image capture use case and attach button click listener
         val imageCapture = ImageCapture(imageCaptureConfig)
         imageCaptureButton.setOnClickListener {
+            deviceOrientation.takeSnapshotAndDisable()
             imageCapture.takePicture(
                 file,
                 object : ImageCapture.OnImageSavedListener {
--- a/app/src/main/java/org/greenstand/android/TreeTracker/models/StepCounter.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/models/StepCounter.kt
@@ -20,13 +20,18 @@ class StepCounter(
     private val stepCounter = sensorManager.getDefaultSensor(Sensor.TYPE_STEP_COUNTER)
     private val stepCountEventListener = StepCountEventListener()
 
+    private val BASE_KEY = PrefKeys.SESSION + PrefKey("steps")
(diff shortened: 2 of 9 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/activities/ImageCaptureActivity.kt
```diff
@@ -9,20 +9,27 @@ import android.view.TextureView
 import android.widget.ImageButton
 import android.widget.TextView
 import androidx.appcompat.app.AppCompatActivity
-import androidx.camera.core.*
+import androidx.camera.core.CameraX
```
arunbakt, line 12: The following explicit imports are ktlint enforced. Please ignore

app/src/main/java/org/greenstand/android/TreeTracker/activities/ImageCaptureActivity.kt
```diff
@@ -71,6 +78,7 @@ class ImageCaptureActivity : AppCompatActivity() {
         // Build the image capture use case and attach button click listener
         val imageCapture = ImageCapture(imageCaptureConfig)
         imageCaptureButton.setOnClickListener {
+            deviceOrientation.takeSnapshotAndDisable()
```
arunbakt, line 81: This call ensures the rotation matrix snapshot is created at the time of the camera button is clicked. Given the rate at which the sensors are generating events, anything else (e.g. at the time of saving the tree) wouldn't accurately present the device orientation at the time of taking the tree's picture.

app/src/main/java/org/greenstand/android/TreeTracker/models/StepCounter.kt
```diff
@@ -41,6 +46,10 @@ class StepCounter(
         sensorManager.unregisterListener(stepCountEventListener)
     }
 
+    fun snapshotAbsoluteStepCountOnTreeCapture() {
```
arunbakt, line 49: Added this function so that the variable absoluteStepCountOnTreeCapture  is modified only via this function and only read directly.

app/src/main/java/org/greenstand/android/TreeTracker/models/StepCounter.kt
```diff
@@ -53,11 +62,4 @@ class StepCounter(
             }
         }
     }
-
-    companion object {
```
arunbakt, line 57: Given there is no intention to expose the keys for accessing the values directly from preferences, I made these private val members of the class. The need isn't there to advocate for a singleton companion class.

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/MapViewModel.kt
```diff
@@ -58,6 +60,7 @@ class MapViewModel constructor(
     fun turnOnTreeCaptureMode() {
         locationDataCapturer.turnOnTreeCaptureMode()
         stepCounter.enable()
+        deviceOrientation.enable()
```
arunbakt, line 63: Tried to have a function in DeviceOrientation the enables, takesSnapshot and disables the listener to see if that helps to avoid this call here. But that model didn't really register any value for the rotation matrix and hence the need for this call separate from the point where the ImageCaptureActivity calls the takeSnapshotAndDisable method.

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/NewTreeViewModel.kt
```diff
@@ -68,7 +70,10 @@ class NewTreeViewModel(
         val deltaSteps = absoluteStepCount - lastStepCountWhenCreatingTree
         newTree.addTreeAttribute(Tree.ABS_STEP_COUNT_KEY, absoluteStepCount.toString())
         newTree.addTreeAttribute(Tree.DELTA_STEP_COUNT_KEY, deltaSteps.toString())
-
+        deviceOrientation.rotationMatrixSnapshot?.let {
```
arunbakt, line 73: I modeled the rotationMatrixSnapshot as a nullable value because the called here has no knowledge whether a invocation of takeSnapshotAndDisable method is called or not in DeviceOrientation. The nullability takes the guessing out.

app/src/main/java/org/greenstand/android/TreeTracker/models/StepCounter.kt
Elforama: The idea behind companion objects is to have 'static' values. There is no need for these values to be tied to the class instance.

Elforama (changes requested):
Looks good, just one minor change.

app/src/main/java/org/greenstand/android/TreeTracker/models/StepCounter.kt
Elforama: Private val makes sense, but we should leave them in a companion object.

arunbakt: I don't get the reason for companion object here. If it were to be accessed/referenced from outside what you said makes sense to me. Absence of that, I don't get it. Could you please point out what I am missing?

arunbakt: @Elforama, I see that you want to keep them as static private variables. Sorry I missed to get it from your earlier comment. Updating it. Sometimes I miss the obvious.

arunbakt (commented):
@Elforama , Updated to restore the companion object in StepCounter. Please review when you have a chance.

Elforama (approved):

Issue #502, opened 2020-09-30
Signup/Signin flow should remove whitespaces


Elforama:
Signup/Signin flow do not remove whitespaces from text input. This is causing issues on the backend when trying to look up a name. We need to remove white spaces

PR #503, opened 2020-10-01
Trim input for signin and signup screens

Elforama:
https://github.com/Greenstand/treetracker-android/issues/502

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/LoginViewModel.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/LoginViewModel.kt
@@ -55,9 +55,9 @@ class LoginViewModel(private val dao: TreeTrackerDAO,
     }
 
     fun updatePhone(phone: String) {
-        val trimmerPhone = phone.trim()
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/LoginViewModel.kt
```diff
@@ -54,8 +55,9 @@ class LoginViewModel(private val dao: TreeTrackerDAO,
     }
 
     fun updatePhone(phone: String) {
-        if (Validation.isValidPhoneNumber(phone)) {
-            this.phone = phone
+        val trimmerPhone = phone.trim()
```
arunbakt, line 58: I believe you meant `trimmedPhone`

arunbakt (approved):

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/LoginViewModel.kt
Elforama: Thanks

PR #504, opened 2020-10-01
Add Kasiki Hai GPS button

Elforama:
Adds GPS button to NewTreeFragment.

Disables convergence on Map screen and moves it to when the new GPS button is pressed. This is all based on feature flags.


app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/NewTreeViewModel.kt
```diff
@@ -38,6 +38,9 @@ class NewTreeViewModel(
 
     val isTreeHeightEnabled = FeatureFlags.TREE_HEIGHT_FEATURE_ENABLED
 
+    var isNextButtonActive = !isDbhEnabled
```
arunbakt, line 41: What is this boolean field for? I couldn't see its usage.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/NewTreeFragment.kt
```diff
@@ -103,6 +107,24 @@ class NewTreeFragment :
                     fragmentNewTreeDBH.text.toString())
             }
         }
+
+        fragmentNewTreeGPS.setOnClickListener {
+            lifecycleScope.launch {
+                progressBar.visibility = View.VISIBLE
+                progressBar.setOnClickListener {
+                    // Do nothing. We want to intercept taps on screen while loading is shown
```
arunbakt, line 115: What would happen if we don't set the listener here?

app/src/main/java/org/greenstand/android/TreeTracker/fragments/NewTreeFragment.kt
```diff
@@ -103,6 +107,24 @@ class NewTreeFragment :
...
+                vm.waitForConvergence()
+                progressBar.visibility = View.GONE
+                fragmentNewTreeSave.isEnabled = true
+                progressBar.setOnClickListener(null)
+            }
+        }
+    }
+
+    override fun onStop() {
+        vm.stopTreeCapture()
```
arunbakt, line 126: When I checked out your branch for the review, I see this method `stopTreeCapture` is missing and results in compilationError. Can you check?

arunbakt (approved):
Looks good to me. I got an compilation error when checking out this branch and I have pointed out the place where it happens in one of the comments. Could you please see why that is?

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/NewTreeViewModel.kt
Elforama: I forgot to push my last changes. It should compile now

app/src/main/java/org/greenstand/android/TreeTracker/fragments/NewTreeFragment.kt
Elforama: If we don't set the click listener than the view will not intercept clicks. This is covering the whole screen to avoid someone from doing something that can mess things up.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/NewTreeFragment.kt
Elforama: Fixed

PR #508, opened 2020-10-03
Cleanup new tree request

Elforama:
Removed default values to make sure we always set the fields.
Made some fields non null

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/api/models/requests/NewTreeRequest.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/api/models/requests/NewTreeRequest.kt
@@ -6,7 +6,7 @@ data class NewTreeRequest(
     @SerializedName("user_id")
     val userId: Int = 0,
     @SerializedName("uuid")
-    val uuid: String? = null,
+    val uuid: String,
     @SerializedName("lat")
     val lat: Double = 0.toDouble(),
     @SerializedName("lon")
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/api/models/requests/NewTreeRequest.kt
```diff
@@ -4,31 +4,31 @@ import com.google.gson.annotations.SerializedName
...
+    val uuid: String,
     @SerializedName("lat")
-    val lat: Double = 0.toDouble(),
+    val lat: Double,
     @SerializedName("lon")
-    val lon: Double = 0.toDouble(),
+    val lon: Double,
     @SerializedName("gps_accuracy")
-    val gpsAccuracy: Int = 0,
+    val gpsAccuracy: Int,
```
arunbakt, line 15: I believe gpsAccuracy is not relevant anymore given the gps convergence work, if I remember the conversation with @deepwinter on this correctly.  The references in the code  to gspAccuracy is still pending to be removed. Shouldn't we remove it from the NewTreeRequest too instead of making it a required field?

arunbakt (changes requested):
Left a comment about gpsAccuracy. I believe we should remove it from the NewTreeRequest altogether since it is no longer relevant given we are using gps convergence, or atleast leave it as an optional field until we completely remove the references to it from the codebase.

Issue #505, opened 2020-10-01
Update convergence to unconverge if GPS data wanders after a capture 


arunbakt:
Currently the GPS convergence once attained remains as converged even if the GPS data wanders within the time the user saves  the capture.

We want the GPS location data to be marked as un-converged if it were to wander before the user hits save. In this scenario, the converged location data will correspond to the location immediately after the convergence is attained and is mostly likely associated to the picture taken after the convergence.  But subsequent locations may un-converge since the planter may move from the spot and hits save at a later point in time from different location than when he captured the picture.

arunbakt:
Issue addressed in  https://github.com/Greenstand/treetracker-android/pull/509 and merged

PR #509, opened 2020-10-05
Allow convergence to un-converge within a  tree capture session

arunbakt:
This change addresses the issue reported in https://github.com/Greenstand/treetracker-android/issues/505.

Convergence computation is continued even after convergence is attained to detect any changes and recorded until the tree capture ends on saving the tree.

Testing done on both greenstand and justdigit flavor

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
@@ -151,8 +151,12 @@ class MainActivity : AppCompatActivity(), ActivityCompat.OnRequestPermissionsRes
 
     public override fun onDestroy() {
         super.onDestroy()
+        locationDataCapturer.turnOffTreeCaptureMode()
         lifecycle.removeObserver(stepCounter)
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
```diff
@@ -151,8 +151,12 @@ class MainActivity : AppCompatActivity(), ActivityCompat.OnRequestPermissionsRes
 
     public override fun onDestroy() {
         super.onDestroy()
+        locationDataCapturer.turnOffTreeCaptureMode()
         lifecycle.removeObserver(stepCounter)
         lifecycle.removeObserver(deviceOrientation)
+        // This is address the use case when the app screen is exited when in
+        // in a tree capture mode
+        locationDataCapturer.turnOffTreeCaptureMode()
```
Elforama, line 159: turnOffTreeCaptureMode() is called twice here

Elforama (changes requested):
Looks good! Just a minor comment.

Elforama (approved):

PR #516, opened 2020-10-12
Fix failing maps screen by creating Google Maps API keys

arunbakt:
The change is to generate new API key for Google Maps from treetracker project in Google APIs console using the id android@greenstand.org.

I have created separate API keys for debug and prod environment restrict each of their access by SHA1 signature in the console as per the document steps.

Tested and checked the maps screen works now.


app/build.gradle
```diff
@@ -23,21 +23,12 @@ apply plugin: 'com.google.firebase.crashlytics'
 
 def loadExtraProperties(String fileName) {
     def props = new Properties()
-    if (System.getenv('CI_BITRISE') == "true") {
```
arunbakt, line 26: Removed to avoid confusion  by all these dead paths that are no longer used

app/build.gradle
```diff
@@ -73,6 +64,8 @@ android {
         buildConfigField 'Boolean', 'USE_AWS_S3', "false"
         buildConfigField 'Boolean', 'ORG_LINK', "false"
 
+        resValue "string", "maps_api_key", "\"${maps_api_key}\""
```
arunbakt, line 67: Recommended way to set the API key in the manifest file via resources. This also allows to vary the api key between test or prod environment.

Elforama (approved):

PR #517, opened 2020-10-13
Config lock screen to dynamically control location update parameters

arunbakt:
The change has ability to see and modify parameters that control 

- how frequently location updates can be captured 
- the min distance required between location updates
- the convergence timeout threshold
- the longitude and latitude convergence threshold value 
- the minimum required data size for computing convergence 

A UI screen for this is part of the change and is hidden in  the following link `About -> build version`
You will require a password to get to the config screen (please DM for password in Slack).  
Saving a modified value will effect the convergence and location updates to behave accordingly.

I made my best effort to create the UI for config screen with just using constraint layout. I believe this meets the needs of
the few users who would require this to fiddle the parameters while using Treetracker app. But I am open to making the UI further refined.


app/src/main/java/org/greenstand/android/TreeTracker/activities/ImageCaptureActivity.kt
```diff
@@ -118,6 +118,11 @@ class ImageCaptureActivity : AppCompatActivity() {
         CameraX.bindToLifecycle(this, preview, imageCapture)
     }
 
+    override fun onBackPressed() {
+        super.onBackPressed()
+        deviceOrientation.disable()
```
arunbakt, line 123: This is to avoid the scenario where device orientation data is continued to be captured when a back button is pressed from within the ImageCaptureActivity.

arunbakt: It is unrelated here but found this while testing the app for this feature.  A minor fix and didn't want to raise a separate issue for resolving this.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/AboutFragment.kt
```diff
@@ -1,46 +1,62 @@
...
     }
 
     override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
         activity?.toolbarTitle?.setText(R.string.information)
         (activity as AppCompatActivity).supportActionBar!!.setDisplayHomeAsUpEnabled(true)
 
-
-        fragmentAboutVersionCode.text = "${getString(R.string.build_version_title)} ${BuildConfig.VERSION_CODE}"
-        fragmentAboutVersionName.text = "${getString(R.string.tree_tracker_title)} ${BuildConfig.VERSION_NAME}"
+        fragmentAboutVersionCode.text = "${getString(R.string.build_version_title)}" +
```
arunbakt, line 38: Ignore the changes spanning lines 38 to 42. Formatting related change to confirm to ktlint

app/src/main/java/org/greenstand/android/TreeTracker/fragments/AboutFragment.kt
```diff
@@ -1,46 +1,62 @@
...
+                " ${BuildConfig.VERSION_CODE}"
+        fragmentAboutVersionName.text = "${getString(R.string.tree_tracker_title)}" +
+                " ${BuildConfig.VERSION_NAME}"
+
+        fragmentAboutVersionCode.setOnTouchListener { v, event ->
```
arunbakt, line 43: The entry point to get to the config lock screen.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/AboutFragment.kt
```diff
@@ -1,46 +1,62 @@
...
+            if (view.findNavController().currentDestination?.id == R.id.aboutFragment) {
+                findNavController()
+                    .navigate(AboutFragmentDirections.actionAboutFragmentToAdminLoginFragment())
+            }
+            true
+        }
     }
 
     override fun onClick(v: View) {
-        v.performHapticFeedback(HapticFeedbackConstants.VIRTUAL_KEY, HapticFeedbackConstants.FLAG_IGNORE_GLOBAL_SETTING)
```
arunbakt, line 39: Ignore ktlint formatting related change

app/src/main/java/org/greenstand/android/TreeTracker/models/Locations.kt
```diff
@@ -297,19 +317,17 @@ class Convergence(val locations: List<Location>) {
         newValue: Double
     ): ConvergenceStats {
         val newMean = currentStats.mean -
-                        (replacingValue / CONVERGENCE_DATA_SIZE) +
-                        (newValue / CONVERGENCE_DATA_SIZE)
+                        (replacingValue / locations.size) +
```
arunbakt, line 320: Given Convergence is passed a list of data named `locations`, the size is already fixed. Convergence wouldn't be created until the required convergence data size is reached  by the caller. Hence it makes sense to use the size of `locations` to do the computation and this also eliminates the need to inject configuration into this object primarily used for computation. The caller instantiating this object is responsible for ensuring the correct size of data is passed for computation and Convergence is only concerned with the given data and its computation.

app/src/test/java/org/greenstand/android/TreeTracker/models/ConvergenceTest.kt
```diff
@@ -52,17 +51,6 @@ class ConvergenceTest {
         convergence = Convergence(locations.subList(0, 5))
     }
 
-    @Test
-    fun noConvergenceUntilThresholdDataSize() {
```
arunbakt, line 56: This test is no longer is relevant for Convergence object given the comment I left for the change in Convergence object above. The responsibility of passing the data with the required size is with the caller.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/ConfigFragment.kt
```diff
@@ -0,0 +1,73 @@
...
+import kotlinx.android.synthetic.main.fragment_config.lonStdDevThresholdVal
+import kotlinx.android.synthetic.main.fragment_config.minDisBtwnUpdatesVal
+import kotlinx.android.synthetic.main.fragment_config.minTimeBtwnUpdatesVal
+import kotlinx.android.synthetic.main.fragment_config.saveConfigButton
+import org.greenstand.android.TreeTracker.R
+import org.greenstand.android.TreeTracker.models.LocationDataConfig
+import org.greenstand.android.TreeTracker.viewmodels.ConfigViewModel
+import org.koin.android.ext.android.inject
+
+class ConfigFragment : Fragment() {
```
Elforama, line 25: For this config screen, I'm wondering if it would be worth making a reusable view that lets us update a value, then we specify a list of PrefKeys (with some additional data) in the ViewModel that will populate the screen. Editing a value can automatically update that PrefKey value.

This would allow us to add more configurable values by simply adding a PrefKey to the list.

Another option could be to use https://developer.android.com/reference/androidx/preference/package-summary

app/src/main/java/org/greenstand/android/TreeTracker/models/Configuration.kt
```diff
@@ -0,0 +1,37 @@
...
+
+    private val LOCATION_DATA_CONFIG_KEY: PrefKey = PrefKeys.SYSTEM_SETTINGS +
+            PrefKey("location-data-config")
+    private val gson = GsonBuilder().serializeNulls().create()
+    lateinit var locationDataConfig: LocationDataConfig
+        private set
+
+    init {
+        locationDataConfig = LocationDataConfig() // Initialize with default value
+        val locationDataConfigString = preferences.getString(LOCATION_DATA_CONFIG_KEY)
```
Elforama, line 18: There doesn't seem to be a nice for this local value.

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/ConfigViewModel.kt
```diff
@@ -0,0 +1,29 @@
...
+
+class ConfigViewModel(
+    private val configuration: Configuration,
+    private val locationUpdateManager: LocationUpdateManager
+) : ViewModel() {
+
+    private val locationConfigLiveData = MutableLiveData<LocationDataConfig>()
+
+    fun getLocationDataConfig(): LiveData<LocationDataConfig> {
+        locationConfigLiveData.value = configuration.locationDataConfig
```
Elforama, line 18: Looks like this can be replace with a call to 'updateLocationDataConfig()'

Elforama (changes requested):
Looking good. Made some minor comments and one large suggestion.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/AboutFragment.kt
```diff
@@ -1,46 +1,62 @@
...
-
-        fragmentAboutVersionCode.text = "${getString(R.string.build_version_title)} ${BuildConfig.VERSION_CODE}"
-        fragmentAboutVersionName.text = "${getString(R.string.tree_tracker_title)} ${BuildConfig.VERSION_NAME}"
+        fragmentAboutVersionCode.text = "${getString(R.string.build_version_title)}" +
+                " ${BuildConfig.VERSION_CODE}"
+        fragmentAboutVersionName.text = "${getString(R.string.tree_tracker_title)}" +
+                " ${BuildConfig.VERSION_NAME}"
+
+        fragmentAboutVersionCode.setOnTouchListener { v, event ->
+            if (view.findNavController().currentDestination?.id == R.id.aboutFragment) {
```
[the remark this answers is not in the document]
arunbakt: This is for the case when the user double taps the link and the navigator has already used the path to navigate on the first touch and the second touch causes the app to crash as there is no known path in the now new screen. Similar to the very first bug I worked on when I started.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/ConfigFragment.kt
```diff
@@ -0,0 +1,73 @@
...
+
+    private val configViewModel by inject<ConfigViewModel>()
+
+    override fun onCreate(savedInstanceState: Bundle?) {
+        super.onCreate(savedInstanceState)
+        setHasOptionsMenu(true)
```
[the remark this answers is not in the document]
arunbakt: Will remove it, I don't see a need for a menu here

app/src/main/java/org/greenstand/android/TreeTracker/fragments/ConfigFragment.kt
arunbakt: I wasn't aware and familiar with this package and its uses, thank you for pointing it. I would agree it is worth to look at this in a subsequent iteration.  I prefer to have the current work go out to make use of this feature for now though.

I too thought about having a dynamic way to produce the configuration screen where specifying a PrefKey would automatically generate the UI by iterating on all the keys and values from the reads. But this would soon become complex since updating one Prefkey at a time would have to trigger a call back to effect its changed value at its point of use (instead of refreshing a cached config object).  Also,  in this specific use case, since location updates causes the config values to be used every second, it is better not to read from preferences directly but rather from a cached value. Given all this, I went the current implementation.  

I am sure there is room for improvement as we add more config objects under here and we could revisit this and examine preference screens.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/ConfigFragment.kt
```diff
@@ -0,0 +1,73 @@
...
+    }
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View? {
+        activity?.toolbarTitle?.setText(R.string.config_parameters)
+        (activity as AppCompatActivity).supportActionBar!!.setDisplayHomeAsUpEnabled(true)
```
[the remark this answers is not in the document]
arunbakt: I was able to change to `requireActivity().toolbarTitle.setText(R.string.config_parameters)` that removed the need to use `?`.

But how can I avoid using `!!` here
        `(activity as AppCompatActivity).supportActionBar!!.setDisplayHomeAsUpEnabled(true)`

Could you specify the preferred way?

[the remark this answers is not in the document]
arunbakt: Replaced 
`(activity as AppCompatActivity).supportActionBar!!.setDisplayHomeAsUpEnabled(true)`
 with 
`(requireActivity() as AppCompatActivity).supportActionBar?.setDisplayHomeAsUpEnabled(true)`

app/src/main/java/org/greenstand/android/TreeTracker/models/Configuration.kt
```diff
@@ -0,0 +1,37 @@
...
+import com.google.gson.GsonBuilder
+import org.greenstand.android.TreeTracker.preferences.PrefKey
+import org.greenstand.android.TreeTracker.preferences.PrefKeys
+import org.greenstand.android.TreeTracker.preferences.Preferences
+
+class Configuration(private val preferences: Preferences) {
+
+    private val LOCATION_DATA_CONFIG_KEY: PrefKey = PrefKeys.SYSTEM_SETTINGS +
+            PrefKey("location-data-config")
+    private val gson = GsonBuilder().serializeNulls().create()
```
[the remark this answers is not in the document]
arunbakt: There were several classes and some instantiating Gson for every execution. Made the change to have Gson injected in all the classes

app/src/main/java/org/greenstand/android/TreeTracker/models/Configuration.kt
arunbakt: I am not sure what your observation is. What is happening here is `LocationDataConfig` takes the constructor default values for the config parameters and is overriden by the values from preferences if a value for `LOCATION_DATA_CONFIG_KEY` is found. Otherwise retains the default constructor values. Required the very first time the app is installed or when the user hasn't fiddled with the config parameters.

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/ConfigViewModel.kt
arunbakt: Sorry, I don't understand your observation here.

getLocationDataConfig returns LiveData with most current value for locationDataConfig and used by UI code interested in the changed value.  We would use updateLocationDataConfig only to update the locationDataConfig in Configuration and to persist the changes.

arunbakt (commented):
@Elforama , Thank you for the review. Updated the PR with some of the changes you have requested and left comments on others. Please take a  look when you get a chance.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/ConfigFragment.kt
Elforama: Android preferences creates its own internal cache, so it's safe to read directly. But yes, let's make any changes in another iteration.

app/src/main/java/org/greenstand/android/TreeTracker/models/Configuration.kt
Elforama: Sorry, typo. Ignore this

Elforama (approved):
Nice!

PR #525, opened 2020-10-24
Release 1.4 alpha

arunbakt:
Bring in the following changes to release 1.4-alpha

- fix ignored image resize on tree captures
- device orientation data, 
- location data config lock screen


arunbakt:
Closing since the unit test check is stuck even though it passes in the private repo build after updates.

PR #533, opened 2020-10-26
Offset -90 degree image rotation that happens on resizing

arunbakt:
Offsets the -90 degree image rotation that happens on resize

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/utilities/ImageUtils.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/utilities/ImageUtils.kt
@@ -378,10 +378,11 @@ object ImageUtils {
 
         /* Decode the JPEG file into a Bitmap */
         val bitmap = BitmapFactory.decodeFile(path, bmOptions)
-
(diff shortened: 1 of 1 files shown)


deepwinter:
Does the rotation happen on ALL phones, or just particular phones?

arunbakt:
It happens on my s8 galaxy. Not sure about other phones.

deepwinter:
There is a common bug in Android since the very beginning of Android where SOME phones mis-identify the orientation of an image, and produce rotated images.  This does not occur on ALL phones.  So we need to make sure we understand how to know the correct orientation of the image (perhaps manually checking the aspect ratio) to make sure we are not introducing rotation in some cases where we think we are fixing it, but are actually correct.   Are you familiar with this classic issue?

deepwinter:
I tested on 3 android phones and they all show the rotation bug (Pixel2, Galaxy A20s, Y9)

deepwinter:
Also tested on an old, terrible Android, J3v6.  Also rotates.  So that's good news, this might not be different across devices.

PR #547, opened 2020-11-03
Make the URLs on the about page function as links

arunbakt:
Make the URLs and email info clickable in the about page in treetracker app.


sebastiangaertner:
Could we update one of the strings in https://github.com/Greenstand/treetracker-android/blob/master/app/src/justdiggit/res/values/strings.xml to see how it is done ?

Issue #548, opened 2020-11-04
Error with back key in justdiggit version


arunbakt:
Issue as reported by user.

"Error with the back key. On pressing the app back key while you have reached the stage of estimating tree height, it takes you back to the beginning (add a tree) instead of the subsequent step of recording GPS coordinates. Then, on starting over the process the app crashes and stops working."

Was able to reproduce this in my phone. From the screen where we estimate the tree height, pressing back key on the phone takes the user to the camera screen. On taking the picture, the app crashes.

PR #553, opened 2020-11-05
Back key on tree height fragment to restart the tree capture flow

arunbakt:
This change fixes the error when back key is pressed when the user is at the tree height fragment screen.
The error behavior is documented here  https://github.com/Greenstand/treetracker-android/issues/548.

The standard navigation backstack pop takes the screen to the camera view instead of the New tree fragment.
This is probably due to the programmatic navigation to the camera view when new tree fragment is loaded.
User proceeding from that points throws a navigation error since the navgraph doesn't have the ImageCaptureActivity in its
design.

Given the existing design, and the states involved in new tree fragment (tree_uuid, location, step count data), the reliable
fix for the upcoming release would take the user to the map screen if user presses the back navigation. This way the app crash
is prevented while making sure the data integrity is maintained.

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/fragments/TreeHeightFragment.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/fragments/TreeHeightFragment.kt
@@ -5,10 +5,10 @@ import android.view.LayoutInflater
 import android.view.View
 import android.view.ViewGroup
 import android.widget.Toast
+import androidx.activity.addCallback
 import androidx.constraintlayout.widget.ConstraintLayout
 import androidx.constraintlayout.widget.ConstraintSet
 import androidx.fragment.app.Fragment
-import androidx.fragment.app.FragmentManager
 import androidx.lifecycle.Observer
 import androidx.navigation.fragment.findNavController
 import androidx.navigation.fragment.navArgs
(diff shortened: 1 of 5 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/fragments/TreeHeightFragment.kt
```diff
@@ -59,22 +62,35 @@ class TreeHeightFragment : Fragment() {
             vm.saveNewTree()
         }
 
-        vm.toastMessagesLiveData().observe(this, Observer {
-            Toast.makeText(context, it, Toast.LENGTH_SHORT).show()
-        })
+        vm.toastMessagesLiveData().observe(
+            this,
```
arunbakt, line 66: ktlint formatting, please ignore

app/src/main/java/org/greenstand/android/TreeTracker/fragments/TreeHeightFragment.kt
```diff
@@ -59,22 +62,35 @@ class TreeHeightFragment : Fragment() {
...
+            Observer {
+                findNavController().popBackStack(R.id.mapsFragment, false)
+            }
+        )
 
-        vm.onFinishedLiveData().observe(this, Observer {
-            findNavController().popBackStack(R.id.mapsFragment, false)
-            fragmentManager?.popBackStack(ValueHelper.NEW_TREE_FRAGMENT, FragmentManager.POP_BACK_STACK_INCLUSIVE)
-        })
+        vm.onEnableButtonLiveData().observe(
```
arunbakt, line 79: ktlint formatting, please ignore

app/src/main/java/org/greenstand/android/TreeTracker/fragments/TreeHeightFragment.kt
```diff
@@ -59,22 +62,35 @@ class TreeHeightFragment : Fragment() {
...
+            this,
+            Observer {
+                save_tree_height.isEnabled = it
+            }
+        )
 
-        vm.onEnableButtonLiveData().observe(this, Observer {
-            save_tree_height.isEnabled = it
-        })
+        requireActivity().onBackPressedDispatcher.addCallback(this) {
```
arunbakt, line 86: This is change that traps the event when back button is pressed and takes the user to the maps screen.

app/src/main/java/org/greenstand/android/TreeTracker/fragments/TreeHeightFragment.kt
```diff
@@ -133,8 +149,14 @@ class TreeHeightFragment : Fragment() {
 
                 ConstraintSet().apply {
                     clone(view)
-                    connect(R.id.floating_button, ConstraintSet.TOP, ConstraintSet.PARENT_ID, ConstraintSet.TOP, 32)
-                    connect(R.id.floating_button, ConstraintSet.BOTTOM, R.id.save_tree_height, ConstraintSet.TOP, 32)
+                    connect(
```
arunbakt, line 152: ktlint formatting, please ignore

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/MapViewModel.kt
```diff
@@ -54,11 +54,10 @@ class MapViewModel constructor(
     }
 
     fun turnOnTreeCaptureMode() {
+        deviceOrientation.enable()
```
arunbakt, line 57: device orientation is captured and disabled in the image capture activity and hence it has to be enabled for both kasiki hai flow and greenstand flow

Elforama (approved):

Issue #554, opened 2020-11-11
Some planter ids are duplicated on ingestion


deepwinter:
Some registration payload do not reach the cloud.  This cascades into a poorly handled edge case in the data pipeline code that causes a new planter record to be created for each tree uploaded.  This has only affected a very limited number of uploads so far and appears to be pretty rare.

The issue is also happening reliably on the test environment.  Next step is to understand the usage conditions that create the problem - code tracing did not reveal any obvious oversights.

arunbakt:
I investigated this issue for planter identifier 099526317 and 0772946  whose registration were missing the database.

I downloaded the files uploaded on 11.10.2020 from s3 and investigated files that had  relatively smaller size(3digit sizes) assuming the registrations will have smaller sized files. I only see the phone identifier in the tree capture file and no registration data turned up for the identifier 099526317  in the set I downloaded. 


But for the user with identifier 0772946 I found the registration record in the upload file. I suspect this is probably a upstream processing issue during ingestion and not in the device. Closing the issue, please re-open if you find new info and suspect this is still the device.

Here is the file name and its related record.
`2020.11.10.14.36.13_ca90c48e-7da9-4694-9aad-640fdf13b1aa_87aed0ec1f3071a92f8ff088041660de:{"devices":[{"androidRelease":"5.0","app_version":"1.3.0","brand":"samsung","device":"hlte","device_identifier":"868cb787efd91c46","hardware":"qcom","instanceId":"
erlzT4oQ-kg","manufacturer":"samsung","model":"SM-N9005","serial":"91ef9603","androidSdkVersion":21,"app_build":120}],"registrations":[{"device_identifier":"868cb787efd91c46","first_name":"Conteh","last_name":"Rosaline","lat":8.47560622,"lon":-13.218977
1,"organization":"FCCAFW ","planter_identifier":"0772946"}]}`

sebastiangaertner:
Thank you. I will keep an eye on it. 
There was some more identifiers though: 08855006 074897309 088464930 07684570

arunbakt:
Reopening the issue since this could be because of bug during ingestion (off the device) and upload registrations (on the device).
Some of the identifiers 07684570, 0772946 do exist in the uploaded registration files but others are not present.

sebastiangaertner:
Its still happening. We have a case now with userid 3021
3020
3019
3018
3017
3016
3015
3014
3013
3012
3011
3010
3009
3008
3007
3006
3005
3004
3003
[...]

PR #556, opened 2020-11-15
Fix for planter registrations failing to be uploaded to s3

arunbakt:
This PR fixes the issue reported in https://github.com/Greenstand/treetracker-android/issues/554 along with one other issue discovered while debugging the reported issue.

1. The local_photo_path value in planter_check_in entity were nullified eagerly regardless of whether they are uploaded to the s3 bucket or not in `DeleteOldPlanterImagesUseCase.kt`.  This had a effect when syncing a bunch of registrations on a slow network   while the planter switches to register a new user.  The update_planters job when entering the delete old planter images routine, fetches all the checkin records (including the new user) to purge the photos without considering if the photo is uploaded to s3. This on a subsequent sync for the new planter registered throws a null pointer error while processing planter checkin thereby failing the registration upload.

 The fix moved is to move the registration sync to happen first along with adding filters to only consider purging local images of planters who have their images sycned to s3 in the checkin/deletion process.

2. Repeated planter check-ins were left unprocessed i.e. images/locations were not uploaded and were left unprocessed in the db. Only the very first check-in on registration was processed. The fix is to use an appropriate query to fetch the correct set of planter checkins (required an index to be added to planter_checkin entity).

### DIFF
--- a/app/schemas/org.greenstand.android.TreeTracker.database.AppDatabase/7.json
+++ b/app/schemas/org.greenstand.android.TreeTracker.database.AppDatabase/7.json
@@ -1,0 +1,430 @@
+{
+  "formatVersion": 1,
+  "database": {
+    "version": 7,
+    "identityHash": "e3f96404ac712f0618eaf04e8251407d",
+    "entities": [
+      {
+        "tableName": "planter_check_in",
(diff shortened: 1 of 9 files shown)


sebastiangaertner:
Well spotted! We have been getting more of those duplicates since two days ago.

Elforama (approved):

PR #559, opened 2020-11-22
557 upload bundle image url with registrations

Elforama:
Adds a users "registration image" to the registration bundle upload

This required the CheckIn image uploads to happen before the registration uploads

This PR also consolidated all the user upload logic into one class to simplify things


app/src/main/java/org/greenstand/android/TreeTracker/di/AppModule.kt
```diff
@@ -12,13 +12,7 @@ import com.google.gson.GsonBuilder
...
 import org.greenstand.android.TreeTracker.api.ObjectStorageClient
 import org.greenstand.android.TreeTracker.background.SyncNotificationManager
-import org.greenstand.android.TreeTracker.models.Configuration
-import org.greenstand.android.TreeTracker.models.DeviceOrientation
-import org.greenstand.android.TreeTracker.models.LanguageSwitcher
-import org.greenstand.android.TreeTracker.models.LocationDataCapturer
-import org.greenstand.android.TreeTracker.models.LocationUpdateManager
-import org.greenstand.android.TreeTracker.models.StepCounter
-import org.greenstand.android.TreeTracker.models.User
+import org.greenstand.android.TreeTracker.models.*
```
arunbakt, line 15: Please restore the individual imports to avoid triggering ktlint errors for others.

arunbakt (changes requested):
@Elforama ,  

Left a comment to restore  explicit imports in AppModule.kt  to avoid triggering the error for others. And another  about using Dispatchers.IO in-place of Dispatchers.Default.

Otherwise the change looks good.

arunbakt (approved):

PR #560, opened 2020-11-23
Avoid hardcoding language into Kasiki Hai flavor

Elforama:
Moved swahili translation into swahili folder.
Used language switcher to default Kasiki Hia flavor to use the swahili translations


app/src/main/java/org/greenstand/android/TreeTracker/activities/MainActivity.kt
```diff
@@ -25,13 +25,7 @@ import org.greenstand.android.TreeTracker.analytics.Analytics
...
 import org.greenstand.android.TreeTracker.fragments.DataFragment
 import org.greenstand.android.TreeTracker.fragments.MapsFragmentDirections
-import org.greenstand.android.TreeTracker.models.DeviceOrientation
-import org.greenstand.android.TreeTracker.models.FeatureFlags
-import org.greenstand.android.TreeTracker.models.LanguageSwitcher
-import org.greenstand.android.TreeTracker.models.LocationDataCapturer
-import org.greenstand.android.TreeTracker.models.LocationUpdateManager
-import org.greenstand.android.TreeTracker.models.StepCounter
-import org.greenstand.android.TreeTracker.models.User
+import org.greenstand.android.TreeTracker.models.*
```
arunbakt, line 28: If you have the android studio setting for code style to use single name import, we can avoid the '*' imports.

app/src/main/java/org/greenstand/android/TreeTracker/activities/SplashFragment.kt
```diff
@@ -24,6 +26,7 @@ class SplashFragment : Fragment() {
 
     private val user: User by inject()
     private val preferencesMigrator: PreferencesMigrator by inject()
+    private val languageSwitcher: LanguageSwitcher by inject()
```
arunbakt, line 29: I don't see languageSwitcher used in this class, am I missing something?

arunbakt (commented):
Have a question about the use of LanguageSwitcher in SplashFragment and the usual import using '*'.

app/src/main/java/org/greenstand/android/TreeTracker/activities/SplashFragment.kt
Elforama: Good catch

arunbakt (approved):

PR #569, opened 2021-01-31
Cleanup Gradle

Elforama:
Updated gradle version
Added ViewBindings to eventually replace synthetics usage
Removed dependencies that stopped gradle upgrade
Cleaned things up

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -67,6 +67,10 @@ android {
         }
     }
 
+    buildFeatures {
+        viewBinding true
+    }
+
     kotlinOptions {
         jvmTarget = "1.8"
     }
@@ -184,9 +188,6 @@ android {
 
     }
 }
-
-
-
 
 androidExtensions {
     experimental = true
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -6,7 +6,6 @@ buildscript {
 
     dependencies {
         classpath 'io.fabric.tools:gradle:1.28.0'
-        classpath 'com.github.triplet.gradle:play-publisher:1.2.0'
     }
 }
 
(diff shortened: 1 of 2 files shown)


app/build.gradle
```diff
@@ -1,20 +1,7 @@
-buildscript {
-    repositories {
-        maven { url 'https://maven.fabric.io/public' }
-        jcenter()
-    }
-
-    dependencies {
-        classpath 'io.fabric.tools:gradle:1.28.0'
-        classpath 'com.github.triplet.gradle:play-publisher:1.2.0'
```
arunbakt, line 9: Are we not depending on play publisher? What is taking its place?
I assume crashlytics is taking over fabric correct?

arunbakt, line 9: Are we not depending on play publisher? What is taking its place?
I assume crashlytics is taking over fabric correct?

app/build.gradle
```diff
@@ -31,15 +18,10 @@ def loadExtraProperties(String fileName) {
 
 loadExtraProperties("treetracker.keys.properties")
 
-
 android {
     compileSdkVersion 29
     buildToolsVersion '29.0.0'
-    playAccountConfigs {
-        defaultAccountConfig {
-            jsonFile = file('keys.json')
```
arunbakt, line 40: I forgot what this was meant for.

arunbakt (commented):
It is nice that we are upgrading the version on gradle. I guess all the tools you removed are obstructing the upgrade, am I right?

app/build.gradle
Elforama: I'll double check of play publisher, I don't think it's used.

yes, firebase is replacing fabric

app/build.gradle
Elforama: This is for the play console plugin I think.

PR #588, opened 2021-03-01
Compose app scaffold

Elforama:
Setup basic navigation and screens in compose.

Language Selection
Org Selection
Email/Phone Input
Name Input
Dashboard

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/splash/SplashScreen.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/splash/SplashScreen.kt
@@ -1,0 +1,50 @@
+package org.greenstand.android.TreeTracker.splash
+
+import android.widget.Toast
+import androidx.compose.foundation.Image
+import androidx.compose.runtime.Composable
+import androidx.compose.runtime.LaunchedEffect
+import androidx.compose.ui.layout.ContentScale
+import androidx.compose.ui.platform.LocalContext
+import androidx.compose.ui.res.painterResource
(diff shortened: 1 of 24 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/splash/SplashScreen.kt
```diff
@@ -0,0 +1,50 @@
...
+import org.koin.core.context.GlobalContext
+import timber.log.Timber
+
+
+@Composable
+fun SplashScreen(preferencesMigrator: PreferencesMigrator, navController: NavController?) {
+    LaunchedEffect(true) {
+        Timber.tag("BuildVariant").d("build variant: ${BuildConfig.BUILD_TYPE}")
+        preferencesMigrator.migrateIfNeeded()
+        delay(100)
```
thenotoriousrog, line 26: Question: What does this delay do? I've never seen this before. Curious on what it does!

thenotoriousrog (approved):
Looks good!

app/src/main/java/org/greenstand/android/TreeTracker/splash/SplashScreen.kt
Elforama: It's just to show the splash screen for 1 second, but it's .1 second while building out the app

Elforama: It pauses the coroutine.

PR #589, opened 2021-03-06
setup the basic starting point for some models

Elforama:


thenotoriousrog (approved):
Looks good for the starting point

PR #590, opened 2021-03-11
Compose Login Updates [Work In Progress]

thenotoriousrog:
This work converts the LoginFragment to using Compose. I've commented out the code that I am going to delete, but decided to keep it for now until I figure out if my path forward on the views makes sense. 

I've updated the LoginViewModel, the LoginFragment, and created the LoginView in compose. 

The PR is broken but will be fixed after an initial view from the Android Lead @Elforama


app/src/main/java/org/greenstand/android/TreeTracker/login/LoginView.kt
```diff
@@ -0,0 +1,221 @@
...
+        keyboardActions = KeyboardActions(
+            onGo = onActionGo
+        )
+
+    )
+}
+
+@Composable
+private fun LoginButton(
+    emailTextState: MutableState<String>,
```
Elforama, line 205: Is there a reason we don't just pass strings to this method?

app/src/main/java/org/greenstand/android/TreeTracker/login/LoginView.kt
```diff
@@ -0,0 +1,221 @@
...
+
+@Composable
+fun LoginView(
+    viewModel: LoginViewModel,
+    navController: NavController
+) {
+
+    val uiScope = rememberCoroutineScope()
+    val snackBarHostState = remember { SnackbarHostState() }
+    val phoneTextInputState = mutableStateOf("")
```
Elforama, line 52: I'm still new to compose, but shouldn't mutableStateOf() be in a remember block?

app/src/main/java/org/greenstand/android/TreeTracker/login/LoginFragment.kt
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
+    }
+
+    override fun onViewCreated(view: View, savedInstanceState: Bundle?) {
+
+        mainActivity().bindings.toolbarTitle.apply {
+            setText(R.string.greenstand_welcome_text)
+            setTextColor(ContextCompat.getColor(context, R.color.black))
+        }
```
Elforama, line 43: Looks like you took this from the old version. I think we should handle any toolbars (any UI) in compose. There are a few weird practices at the Activity/Fragment level that the legacy code does, so we should feel free to find other solutions.

app/src/main/res/navigation/nav_graph.xml
```diff
@@ -82,7 +82,7 @@
         </fragment>
         <fragment
             android:id="@+id/loginFragment"
-            android:name="org.greenstand.android.TreeTracker.fragments.LoginFragment"
+            android:name="org.greenstand.android.TreeTracker.login.LoginFragment"
```
Elforama, line 85: There is a second navGraph you can use instead of this one. This is for the legacy app flow, I'd like to keep them in parallel until the new UI is complete.

app/src/main/java/org/greenstand/android/TreeTracker/viewmodels/LoginViewModel.kt
```diff
@@ -1,83 +0,0 @@
-package org.greenstand.android.TreeTracker.viewmodels
```
Elforama, line 1: Let's keep all legacy files for now. Currently in master, if you switch the launch Activity to MainActivity the entire old version of the app should be 100% working.

app/src/main/java/org/greenstand/android/TreeTracker/view/TreeTrackerButton.kt
```diff
@@ -0,0 +1,57 @@
...
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.res.stringResource
+import androidx.compose.ui.tooling.preview.Preview
+import androidx.compose.ui.unit.dp
+import org.greenstand.android.TreeTracker.R
+import org.greenstand.android.TreeTracker.utilities.Padding
+
+@Composable
+fun TreeTrackerButton(
```
Elforama, line 20: This looks great!

app/src/main/java/org/greenstand/android/TreeTracker/utilities/CameraHelper.kt
```diff
@@ -14,14 +14,18 @@ object CameraHelper {
     fun takePictureForResult(fragment: Fragment, selfie: Boolean) {
         if (ActivityCompat.checkSelfPermission(
                 fragment.requireContext(),
-                Manifest.permission.CAMERA) != PackageManager.PERMISSION_GRANTED ||
+                Manifest.permission.CAMERA
```
Elforama, line 17: Was al this just reformatting?

app/src/main/java/org/greenstand/android/TreeTracker/login/LoginView.kt
```diff
@@ -0,0 +1,221 @@
...
+                        .height(dimensionResource(id = R.dimen.logo_height)),
+
+                    imageVector = ImageVector.vectorResource(id = R.drawable.logo),
+                    contentDescription = stringResource(id = R.string.greenstand_welcome_text)
+                )
+
+                PhoneInputText(
+                    textInputState = phoneTextInputState,
+                    onActionDone = {
+                        emailFocusRequester.requestFocus()
```
Elforama, line 95: This method of controlling focus reminds me a lot of Flutter.

Elforama:
Overall this looks good. Just remember not to delete or modify the old version of the app. This work should be additive only.

In terms of the screen flow, I'm hoping there is a way for us to contain the whole signup/login flow in a single fragment so we can easily have a single viewmodel handle all the data. Each screen is really just filling in a field or two each. If you look at SignupScreen.kt you'll see what I mean. Although my approach to it there is poor.

What I haven't been able to do yet is intercept the back button to make it swap compose views instead of control the natural backstack. If we can get that working I think it should be simple to handle the login screen flow. Any thoughts/ideas are welcome.

app/src/main/java/org/greenstand/android/TreeTracker/login/LoginView.kt
thenotoriousrog: Yes! Good catch!

app/src/main/java/org/greenstand/android/TreeTracker/login/LoginFragment.kt
thenotoriousrog: Great to know! Will update this appropriately.

app/src/main/res/navigation/nav_graph.xml
thenotoriousrog: Fantastic! I was unaware of this. Will fix

app/src/main/java/org/greenstand/android/TreeTracker/login/LoginView.kt
thenotoriousrog: Yeah I'm not really a fan of it. In all honesty, I just copied the way we handled keyboard actions to move users to another view from my company. For giggles, I'll see if I can find a more intuitive way of doing it.

app/src/main/java/org/greenstand/android/TreeTracker/login/LoginView.kt
thenotoriousrog: Yeah. So Compose updates it's views whenever its `State` changes. By utilizing the `State<T>` interface (`MutableState<T>` implements this interface). So by attaching the email and phone states to the `isEnabled` argument it will ensure that the button enables and disables on it's own without us having to do anything :) it's all handled for us!

thenotoriousrog: I realize that my explanation my have been superfluous! My apologies, I was more so thinking outloud in what I have understood how the UI updates and passing just the text I think wouldn't update the button enabled state in real-time due to the whole state thing

app/src/main/java/org/greenstand/android/TreeTracker/utilities/CameraHelper.kt
thenotoriousrog: Yeah I'm not sure what happened here to be honest. I'll revert the changes in here

thenotoriousrog:
I will definitely look at converting this to one fragment! I'll also take a look at handling the backstack interception

app/src/main/java/org/greenstand/android/TreeTracker/utilities/Padding.kt
```diff
@@ -0,0 +1,18 @@
...
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+
+/**
+ * Simple padding values.
+ *
+ * Useful for supplying padding values
+ * without having to use a Modifier.
+ */
+data class Padding(
```
Elforama, line 12: It looks like padding modifiers have a very nice API. Why don't we just use them instead?

thenotoriousrog: I was looking for an api for that! I will do some digging and use that instead. Much prefer not to reinvent the wheel

thenotoriousrog:
After talking with @Elforama I realize now that converting the Login stuff to compose UI is not needed and that this was unnecessary

PR #591, opened 2021-03-13
Create basic functioning org picker UI

Elforama:


app/src/main/java/org/greenstand/android/TreeTracker/models/Organizations.kt
```diff
@@ -12,19 +16,100 @@ data class Org(
     val captureInputs: List<String>
 )
 
-class Organizations(
+class OrgPickerViewModelPreview: PreviewParameterProvider<OrgPickerViewModel> {
```
thenotoriousrog, line 19: Love this

thenotoriousrog (approved):
Love it! Looks good

PR #592, opened 2021-03-14
basic user select screen

Elforama:
Creates an basic user select screen from the wallet flow

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/userselect/UserSelectFragment.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/userselect/UserSelectFragment.kt
@@ -1,4 +1,4 @@
-package org.greenstand.android.TreeTracker.dashboard
+package org.greenstand.android.TreeTracker.userselect
 
 import android.os.Bundle
 import android.view.LayoutInflater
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/userselect/UserSelectScreen.kt
```diff
@@ -0,0 +1,38 @@
...
+import androidx.compose.runtime.livedata.observeAsState
+import androidx.compose.ui.Alignment
+import androidx.compose.ui.Modifier
+import androidx.compose.ui.unit.dp
+import androidx.navigation.NavController
+import org.greenstand.android.TreeTracker.view.TreeTrackerTheme
+
+@Composable
+fun UserSelectScreen(viewModel: UserSelectViewModel, navController: NavController) {
+    TreeTrackerTheme {
```
thenotoriousrog, line 19: The `TreeTrackerTheme` was already applied in the `UserSelectFragment` so this shouldn't be necessary here. I had issues with applying the same composable twice that led to crashes but that was due to a doubly created `Scaffold` composable. I'm not sure if a doubly set `TreeTrackerTheme` would cause issues but it's technically not needed here

app/src/main/java/org/greenstand/android/TreeTracker/userselect/UserSelectFragment.kt
```diff
@@ -0,0 +1,32 @@
...
+class UserSelectFragment : Fragment() {
+
+    private val viewModel: UserSelectViewModel by viewModel()
+
+    override fun onCreateView(
+        inflater: LayoutInflater,
+        container: ViewGroup?,
+        savedInstanceState: Bundle?
+    ): View {
+        return createCompose(1) {
```
thenotoriousrog, line 23: Are we planning to add compose resIds to an `ids.xml` file later?

thenotoriousrog (changes requested):
Overall this looks great! I'm submitting requesting changes solely so that a notification get's pushed through. My only real big concern here is the doubly set `TreeTrackerTheme` composable. Since it's in the `UserSelectFragment` we don't really need it in the `UserSelectScreen`

app/src/main/java/org/greenstand/android/TreeTracker/userselect/UserSelectFragment.kt
Elforama: Yes, I've been lazy. We should have unique IDs for each screen (unlike now)

PR #595, opened 2021-03-22
Updated behavior to use navigation only for compose [Work in progress]

thenotoriousrog:
In this PR, I've converted our navigation to only use compose. In order to account for the bug in koin compose, I built a `ViewModelFactory` to create all of our viewModels. What's nice is that compose keeps a local `ViewModelStore` and accounts for taking care of any viewModels generated. Koin also does this work in the background. Essentially, the `ViewModelFactory` architecture will give us all of the nice benefits that Koin does with the only added input of a factory instance. 

The factory instance is supplied in Koin as a singleton and is injected inside the `TreeTrackerActivity`. The only drawback is that for a ViewModel to properly be generated for the view to use, an instance of the `ViewModelFactory` must be passed into the view. Thankfully, Android Compose supplies us with the notion of a `CompositionLocalProvider` where we can supply just about anything we want to our views. I created a `LocalViewModelFactory` provider so that every view that's ever created in our navGraph will always be able to access it like so `LocalViewModelFactory.current`. This made things quite clean and clear. 

**Main Points & Nicessities**
* The app works and navigates correctly (from what I've seen, there may be a piece or two I'll have to comeback to, but this PR isn't finished so that works out.
* The app essentially is in 100% compose for the UI
* There's a `viewModelModule` which has all the current viewModels accounted for so that when the bug in Koin compose is fixed we can easily and quickly make the switch to that.
* The `ViewModelFactory` is 100% native and works the way that Koin does under the hood. Or at least, seemingly under the hood.
* The current implementation, with - of course, some code cleanup, is ready, workable, and easy to build off of while keeping the app at 100% compose. 
* When the bug in koin compose is fixed, the only overhead we'd need to go back to relying on Koin would be to change viewModel requests in the compose views to `viewModel()` instead of `viewModel(factory = *suppliedFactory*)` and removing the `TreeTrackerViewModelFactory`.
* This implementation will simplify the code as we won't have to make the switch between compose and non-compose elements. 
* Nearly all previews should be working in the compose elements touched in this build. There was one that wasn't working but I think that was because I needed to restart Android Studio and didn't want to wait for it to resync and build.

I'd love your feedback on this @Elforama. I know this was yet another big commit, but I wanted to suggest an architecture that would allow us to be 100% compose that's not too intrusive nor offensive. I believe the architecture I provided here is a decent solution and will make navigating simpler as well as reduce the amount of code to write. The architecture above is stable, and works in production for the company I work at and is quite easy to maintain and build upon. That being said, that is _my_ opinion and I'm very much interested in what you think about the changes proposed.

Forgive the rather, well, ugly state of the code haha. Before I made it look pretty, I wanted to get the meat and potatoes fleshed out, get feedback, and move on from there.


app/src/main/java/org/greenstand/android/TreeTracker/BaseViewModel.kt
```diff
@@ -0,0 +1,6 @@
+package org.greenstand.android.TreeTracker
+
+import androidx.lifecycle.ViewModel
+import org.koin.core.KoinComponent
+
+abstract class BaseViewModel : ViewModel(), KoinComponent
```
thenotoriousrog, line 6: I did `KoinComponent` here because it'll make it easier for us to inject stuff into our viewModels whenever necessary. While I personally prefer if we can have all of it provided in Koin and then generated in the viewModelFactory so that we get that compile time speed boost, having the KoinComponent will give the implementer extra flexibility when building ViewModels for future features.

app/src/main/java/org/greenstand/android/TreeTracker/activities/TreeTrackerActivity.kt
```diff
@@ -1,34 +1,84 @@
...
+import org.greenstand.android.TreeTracker.languagepicker.LanguageSelectScreen
+import org.greenstand.android.TreeTracker.models.*
+import org.greenstand.android.TreeTracker.signup.SignupScreen
+import org.greenstand.android.TreeTracker.splash.SplashScreen
+import org.greenstand.android.TreeTracker.view.TreeTrackerTheme
 import org.koin.android.ext.android.inject
+import timber.log.Timber
 
-class TreeTrackerActivity : AppCompatActivity() {
+val LocalViewModelFactory = compositionLocalOf<TreeTrackerViewModelFactory> { error("No active ViewModel factory found!") }
```
thenotoriousrog, line 21: This is what I mentioned in the write up. Essentially, this allow the `ViewModelFactory` to exist in the compose runtime and be accessible to all of our views going forward

app/src/main/java/org/greenstand/android/TreeTracker/activities/TreeTrackerActivity.kt
```diff
@@ -1,34 +1,84 @@
...
             languageSwitcher.applyCurrentLanguage(this)
         }
 
-        bindings = TreeTrackerActivityBinding.inflate(layoutInflater)
+        // TODO: Remember to undo these so that the NavigationGraph can still do it's thing
+//        bindings = TreeTrackerActivityBinding.inflate(layoutInflater)
+//        setContentView(bindings.root)
+
+        setContent {
+            CompositionLocalProvider(LocalViewModelFactory provides viewModelFactory) {
```
thenotoriousrog, line 46: All views inside here ensures access to that viewModel factory

app/src/main/java/org/greenstand/android/TreeTracker/activities/TreeTrackerActivity.kt
```diff
@@ -1,34 +1,84 @@
...
+
         if (FeatureFlags.USE_SWAHILI) {
             languageSwitcher.setLanguage(Language.SWAHILI, resources)
         } else {
             languageSwitcher.applyCurrentLanguage(this)
         }
 
-        bindings = TreeTrackerActivityBinding.inflate(layoutInflater)
+        // TODO: Remember to undo these so that the NavigationGraph can still do it's thing
+//        bindings = TreeTrackerActivityBinding.inflate(layoutInflater)
```
thenotoriousrog, line 42: This is no longer needed as we now have a compose `NavGraph`

app/src/main/java/org/greenstand/android/TreeTracker/models/TreeTrackerViewModelFactory.kt
```diff
@@ -0,0 +1,29 @@
...
+import androidx.lifecycle.ViewModel
+import androidx.lifecycle.ViewModelProvider
+import java.lang.RuntimeException
+import org.greenstand.android.TreeTracker.dashboard.DashboardViewModel
+import org.greenstand.android.TreeTracker.languagepicker.LanguagePickerViewModel
+import org.greenstand.android.TreeTracker.preferences.PreferencesMigrator
+import org.greenstand.android.TreeTracker.signup.SignupViewModel
+import org.greenstand.android.TreeTracker.splash.SplashScreenViewModel
+
+@Suppress("UNCHECKED_CAST")
```
thenotoriousrog, line 13: I think this suppression can be removed by using `reified` but I can look into that later

Elforama: I don't think reified can be used on classes or override methods

app/src/main/java/org/greenstand/android/TreeTracker/models/TreeTrackerViewModelFactory.kt
```diff
@@ -0,0 +1,29 @@
...
+    private val languageSwitcher: LanguageSwitcher,
+    private val resources: Resources
+) : ViewModelProvider.NewInstanceFactory() {
+
+    override fun <T : ViewModel?> create(modelClass: Class<T>): T {
+        return when {
+            modelClass.isAssignableFrom(SplashScreenViewModel::class.java) -> SplashScreenViewModel(preferencesMigrator) as T
+            modelClass.isAssignableFrom(LanguagePickerViewModel::class.java) -> LanguagePickerViewModel(languageSwitcher, resources) as T
+            modelClass.isAssignableFrom(DashboardViewModel::class.java) -> DashboardViewModel() as T
+            modelClass.isAssignableFrom(SignupViewModel::class.java) -> SignupViewModel() as T
```
Elforama, line 25: Can we rely on koin to inject instances here instead?

app/src/main/java/org/greenstand/android/TreeTracker/splash/SplashScreenPreviewProvider.kt
```diff
@@ -0,0 +1,61 @@
+package org.greenstand.android.TreeTracker.splash
+
+import android.content.SharedPreferences
+import androidx.compose.ui.tooling.preview.PreviewParameterProvider
+import org.greenstand.android.TreeTracker.preferences.Preferences
+import org.greenstand.android.TreeTracker.preferences.PreferencesMigrator
+
+class SplashScreenPreviewProvider : PreviewParameterProvider<SplashScreenViewModel> {
+
+    private val stubSharedPrefs = object : SharedPreferences {
```
Elforama, line 10: We should make a single stub version of this that we can reuse.

app/src/main/java/org/greenstand/android/TreeTracker/BaseViewModel.kt
Elforama: Since that use case hasn't come up, I'd prefer if we didn't overdesign this. As it is now, all our viewmodels use constructor injection which doesn't need this.

app/src/main/java/org/greenstand/android/TreeTracker/di/ViewModelModule.kt
```diff
@@ -0,0 +1,16 @@
+package org.greenstand.android.TreeTracker.di
+
+import org.greenstand.android.TreeTracker.dashboard.DashboardViewModel
+import org.greenstand.android.TreeTracker.languagepicker.LanguagePickerViewModel
+import org.greenstand.android.TreeTracker.signup.SignupViewModel
+import org.greenstand.android.TreeTracker.splash.SplashScreenViewModel
+import org.koin.android.viewmodel.dsl.viewModel
+import org.koin.dsl.module
+
+val viewModelModule = module {
```
Elforama, line 10: This is already done in the app module, but this is cleaner. For now, we can delete this and break up the app module is a different PR

Elforama:
Thanks for doing all this updating. Overall it seems to make sense to use the compose navigation. It seems to make things easier, and no more fragments to worry about. If your company is using it, then it's probably stable enough for us. 

Moving forwards please keep each PR to a single task. Not only is it easier to review, but it's faster to create the PR, and easier to fix rebase issues.

Elforama (changes requested):
Looks like you'll have to rebase and add a couple more screens to the navigation

app/src/main/java/org/greenstand/android/TreeTracker/models/TreeTrackerViewModelFactory.kt
thenotoriousrog: Hmmmm I think so yes! Let me explore that! I think I can implement the `KoinComponent` interface and be able to inject since the viewModels are created in the modules for us already.

app/src/main/java/org/greenstand/android/TreeTracker/splash/SplashScreenPreviewProvider.kt
thenotoriousrog: I was thinking this as well. I noticed I was repeating code and will likely need to repeat this again. I think I'll make a `PreviewProviderUtil` class or something with common objects that be used as stubs

app/src/main/java/org/greenstand/android/TreeTracker/di/ViewModelModule.kt
thenotoriousrog: Ah okay, that works! Maybe I'll do that this weekend for funsies. For this PR I'll move things over and nuke it

thenotoriousrog:
Sweet, can do. Yeah no problem. Going forward things will be much, much simpler.

app/src/main/java/org/greenstand/android/TreeTracker/models/TreeTrackerViewModelFactory.kt
thenotoriousrog: Ah okay dang. I hate this warning, but I haven't figured out how to remove it

app/src/main/java/org/greenstand/android/TreeTracker/BaseViewModel.kt
thenotoriousrog: Works for me. Thanks for that call out! I sometimes need that reminder to dial it back. I can easily remove and keep it the way it is.

thenotoriousrog:
Rebase and merge conflicts fixed. I did however, add a `LocalNavHostController` so we can easily obtain it from our composable views

app/src/main/java/org/greenstand/android/TreeTracker/BaseViewModel.kt
thenotoriousrog: Removed

app/src/main/java/org/greenstand/android/TreeTracker/activities/TreeTrackerActivity.kt
```diff
@@ -27,8 +37,48 @@ class TreeTrackerActivity : AppCompatActivity() {
...
         }
 
-        bindings = TreeTrackerActivityBinding.inflate(layoutInflater)
+        setContent {
+            val navController = rememberNavController()
+
+            CompositionLocalProvider(
+                LocalViewModelFactory provides viewModelFactory,
+                LocalNavHostController provides navController
+            ) {
```
Elforama, line 46: Awesome

thenotoriousrog:
@Elforama Changes fixed, rebase complete, and passes the smoke test. Ready for your second pass

Elforama (approved):

PR #608, opened 2021-04-10
Depth button

Elforama:
Create DepthButton to be used as base button for all buttons.

Created DepthButtonColors which copies the structure from Compose to supply button colors with enabled/disabled colors

Updated AppBar to match design

https://user-images.githubusercontent.com/6439445/114274069-bc91c980-99ea-11eb-9680-7ccaff43e7a6.mp4


app/src/main/java/org/greenstand/android/TreeTracker/view/Buttons.kt
```diff
@@ -43,4 +64,167 @@ fun BoxScope.LanguageButton() {
...
+    modifier: Modifier,
+    isPressed: Boolean,
+    color: Color,
+    shadowColor: Color,
+    depth: Float = 20f,
+) {
+
+    val offsetAnimation: Float by animateFloatAsState(targetValue = if (isPressed) 1f else 0f)
+
+    Canvas(modifier = modifier.fillMaxSize()) {
```
thenotoriousrog, line 157: Fancy! Nice custom work here!!

thenotoriousrog (approved):
Looks good! May be some changes I'll have to make in the signup stuff but that's okay!

Super nice to have these buttons!! Love the work. Real fancy!

PR #609, opened 2021-04-17
Signup flow updates

thenotoriousrog:
This PR completes the Sign-In flow with UI updates only. It also includes a custom `BorderedTextField` to help with matching the UX designs. 

The `SignUpViewModel` doesn't do any data validation right now and will need to be patched up later. 

There's a couple of things that I wasn't able to fix during this PR and will likely need to be patched up down the road. 
1. I tried my best to remove the green underline inside the `BorderedTextField` but for some reason it just wouldn't go away. I was able to get it to go away before but for some reason what I did to remove the underline is no longer working. So when the user clicks on the border it gets rid of the underline, but when the user clicks out of the `BorderedTextField` the underline comes back.
2. I wasn't able to get the user sent to the `Dashboard` after taking a screenshot. I'm not exactly sure why. I did some logging and snooped around the code but was unable to figure out why a result isn't being returned after the user snaps a photo. 
3. There's no saved state in the app as well when transferring between the screens.

**Screen Recording:** https://drive.google.com/file/d/1CDPxKTV6sps43skwzIsdJ4-LwjdmZVgJ/view?usp=sharing


app/src/main/java/org/greenstand/android/TreeTracker/dashboard/DashboardScreen.kt
```diff
@@ -32,7 +32,7 @@ fun DashboardScreen(
     Scaffold(
         topBar = {
             DashboardTopBar(navController)
-        },
+        }
```
Elforama, line 35: Please change your IDE settings to not remove trailing commas

app/src/main/java/org/greenstand/android/TreeTracker/models/NavRoute.kt
```diff
@@ -24,6 +24,10 @@ sealed class NavRoute {
         override val route: String = "signup-flow"
     }
 
+    object NameEntryView : NavRoute() {
+        override val route: String = "signup_flow/nameEntryView"
```
Elforama, line 28: please use "-" instead of  "_"

app/src/main/java/org/greenstand/android/TreeTracker/signup/NameEntryView.kt
```diff
@@ -0,0 +1,99 @@
...
+                    TreeTrackerTextButton(
+                        modifier = Modifier.align(Alignment.Center),
+                        stringRes = R.string.language,
+                        onClick = { navController.navigate(NavRoute.Language.create(isFromTopBar = true)) }
+                    )
+                }
+            )
+        },
+        bottomBar = {
+            Row(
```
Elforama, line 68: We should use the ActionBar for bottombar as well

app/src/main/java/org/greenstand/android/TreeTracker/signup/SignupScreen.kt
```diff
@@ -1,54 +1,42 @@
...
+import org.greenstand.android.TreeTracker.ui.mediumGrey
 import org.greenstand.android.TreeTracker.view.ActionBar
+import org.greenstand.android.TreeTracker.view.BorderedTextField
 import org.greenstand.android.TreeTracker.view.LanguageButton
 
 @Composable
 fun SignupFlow(
     viewModel: SignupViewModel = viewModel(factory = LocalViewModelFactory.current),
 ) {
-    val state by viewModel.state.observeAsState(SignUpState())
```
Elforama, line 29: Can we keep this as "state"?

app/src/main/java/org/greenstand/android/TreeTracker/signup/SignupViewModel.kt
```diff
@@ -7,26 +7,55 @@ import org.greenstand.android.TreeTracker.models.Users
...
 // Dequeue breaks equals so state will not be updated when navigating
 data class SignUpState(
-    val emailPhone: String? = null,
+    val emailText: String = "",
+    val phoneText: String = "",
     val name: String? = null,
     val photoPath: String? = null,
+    val credentialType: CredentialType = CredentialType.Email,
+    val showEmailText: Boolean = true,
+    val showPhoneText: Boolean = false,
```
Elforama, line 16: I take it if one is false, the other is true. In that case we only need one boolean.

app/src/main/java/org/greenstand/android/TreeTracker/view/TreeTrackerTextButton.kt
```diff
@@ -36,7 +36,7 @@ import org.greenstand.android.TreeTracker.activities.LocalNavHostController
 import org.greenstand.android.TreeTracker.models.NavRoute
 
 @Composable
-fun TextButton(
+fun TreeTrackerTextButton(
```
Elforama, line 39: Why the rename?

app/src/main/java/org/greenstand/android/TreeTracker/userselect/UserSelectScreen.kt
```diff
@@ -25,7 +25,9 @@ import org.greenstand.android.TreeTracker.view.LanguageButton
 import org.greenstand.android.TreeTracker.view.TextButton
 
 @Composable
-fun UserSelectScreen(viewModel: UserSelectViewModel = viewModel(factory = LocalViewModelFactory.current)) {
+fun UserSelectScreen(
+    viewModel: UserSelectViewModel = viewModel(factory = LocalViewModelFactory.current)
+) {
```
Elforama, line 30: There are tons of files that have nothing but formatting applied. Please remove all changes from files that didn't have real code changes.

app/src/main/java/org/greenstand/android/TreeTracker/ui/TreeTrackerTheme.kt
```diff
@@ -0,0 +1,24 @@
+package org.greenstand.android.TreeTracker.view
+
+import androidx.compose.material.*
+import androidx.compose.runtime.Composable
+import org.greenstand.android.TreeTracker.ui.*
+
+val Colors = darkColors(
```
Elforama, line 7: There is already a colors file

app/src/main/java/org/greenstand/android/TreeTracker/view/BorderedTextField.kt
```diff
@@ -0,0 +1,42 @@
...
+import androidx.compose.ui.graphics.SolidColor
+import androidx.compose.ui.unit.Dp
+import androidx.compose.ui.unit.dp
+
+@Composable
+fun BorderedTextField(
+    padding: PaddingValues = PaddingValues(0.dp),
+    borderSize: Dp = 0.5.dp,
+    borderColor: Color = Color.White,
+    shape: Shape = RoundedCornerShape(16.dp),
```
Elforama, line 24: Since we want all our text input to look the same we shouldn't expose any of these parts.

app/src/main/java/org/greenstand/android/TreeTracker/ui/Color.kt
```diff
@@ -0,0 +1,24 @@
+package org.greenstand.android.TreeTracker.ui
+
+import androidx.compose.ui.graphics.Color
+
+val colorPrimary = Color(0xFF75B926)
```
Elforama, line 5: There is already a colors file

app/src/main/java/org/greenstand/android/TreeTracker/view/TreeTrackerTextButton.kt
thenotoriousrog: For a more conventional naming since it's custom for our TreeTracker app. I can definitely rename it however.

app/src/main/java/org/greenstand/android/TreeTracker/ui/Color.kt
thenotoriousrog: Yeah I think I did this before the latest changes in master. Will fix this as part of the updates

app/src/main/java/org/greenstand/android/TreeTracker/view/BorderedTextField.kt
thenotoriousrog: Will update!

app/src/main/java/org/greenstand/android/TreeTracker/userselect/UserSelectScreen.kt
thenotoriousrog: I'm not entirely sure why that happened. Super weird. I can definitely make sure that no useless updates are getting pushed through

app/src/main/java/org/greenstand/android/TreeTracker/view/TreeTrackerTextButton.kt
Elforama: I didn't noticed a pre-made TextButton, so let's try to keep TreeTracker out of names. I have done that with our theme and activity, but that was due to simple default names being used already. In fact, I'd like those names improved later in another PR.

app/src/main/java/org/greenstand/android/TreeTracker/userselect/UserSelectScreen.kt
Elforama: Thanks, and if you can't undo it, don't worry about it. After this PR I'd suggest looking into your IDE settings. I'm guessing the moment you open a file the IDE auto formats everything. This is nice if we have a unified formatter, but we don't. That's on the to-do list.

app/src/main/java/org/greenstand/android/TreeTracker/signup/SignupScreen.kt
thenotoriousrog: Sure no problem

app/src/main/java/org/greenstand/android/TreeTracker/signup/NameEntryView.kt
thenotoriousrog: I had this originally with the bottomBar but I guess during the merges it got nuked haha. I can make use of the ActionBar instead

app/src/main/java/org/greenstand/android/TreeTracker/dashboard/DashboardScreen.kt
thenotoriousrog: Done

PR #621, opened 2021-05-02
Signup UI Updates [Nav Fix]

thenotoriousrog:
This PR is the same as the previously reverted SignUp UI Updates PR except this one contains the fixes for the navigation. 

The last PR had an unknown issue with navigation that caused the app to not navigate to the `Dashboard` screen when the user had finished taking a photo. Interestingly enough, the navigation code was identical to what was present in the updated UI, but navigation failed despite working perfectly in master.

This PR contains an upgraded dependency for the compose activity. This rectifies the issue that was preventing the app from reaching the `Dashboard` upon finishing a selfie. 

**Video of feature:** https://drive.google.com/file/d/1qBRivmYzj9bHBi6G95i7Na7iLjjY8o8Q/view?usp=sharing


Elforama:
Fantastic, thanks for getting it working!

Issue #615, opened 2021-04-18
Update User Select screen content to new design


Elforama:
Update the UserSelect screen content in UserSelectScreen.kt to the design found here: https://www.sketch.com/s/a0bbeac0-767a-4873-beef-c6c7ffd5826a/a/4a55aKo

The bottom bar (3 buttons) will be updated in another task. This one only focuses on the scrolling list of users.

This task is focused on updating the UX.

PR #625, opened 2021-05-05
615 Update user select screen

MarcellinoCO:
# #615 Update User Select screen content to new design

This changes updated the UserSelectScreen to the new design provided. Here is the screenshot:

## Some Issues Faced:
1. The Box component for the profile picture is not matching its parent width in some devices (as seen in the screenshot), while in other devices it seems fine.
2. I can't seem to find a way to make the UserButton to wrap its content. I need to hardcode the size with the Modifier. I think it would be nicer to make the button to wrap its content.
3. Need to update 'Plant' icon under user's phone number (the icon can't be exported in the design).

## Where To Go From Here:
1. Need to fetch real user data (profile picture, name, phone, and number of tokens).
2. Update the bottom bar of the layout.


app/src/main/java/org/greenstand/android/TreeTracker/userselect/UserSelectScreen.kt
```diff
@@ -71,22 +81,91 @@ fun UserSelectScreen(
...
+            modifier = Modifier
+                .fillMaxHeight()
+                .fillMaxWidth()
+                .align(Alignment.TopCenter),
             horizontalAlignment = Alignment.CenterHorizontally
         ) {
-            Text("User Select")
-            state.planters.forEach { user ->
+            // User profile picture.
+            Box(  // TODO: Fetch user profile picture.
```
Elforama, line 128: To make this match parent you might be able to use fillMaxWidth and set its aspect ratio to 1:1

app/src/main/java/org/greenstand/android/TreeTracker/userselect/UserSelectScreen.kt
```diff
@@ -71,22 +81,91 @@ fun UserSelectScreen(
...
+            ) {
                 Text(
-                    text = "${user.firstName} ${user.lastName}",
-                    modifier = Modifier
-                        .padding(16.dp)
-                        .clickable {
-                            viewModel.selectUser(user)
-                        }
-                        .background(color = if (state.selectedPlanter?.id == user.id) Color.Gray else Color.White)
+                    text = "Name\nPhone",  // TODO: Fetch user name and phone number.
```
Elforama, line 144: We have access to the name already, let's use that here to keep the functionality.

app/src/main/java/org/greenstand/android/TreeTracker/userselect/UserSelectScreen.kt
```diff
@@ -71,22 +81,91 @@ fun UserSelectScreen(
...
+                ) { viewModel.selectUser(user) }
+            }
+        }
+    }
+}
 
+@Composable
+fun UserButton(
+    user: PlanterInfoEntity,
+    state: UserSelectState,
```
Elforama, line 102: We're only passing in the screens state object to know if it's selected. To decouple the list item from the screen I'd suggest passing in an isSelected boolean instead.

Elforama (changes requested):
I left a few comments to address, but overall this looks great!

I'll write up some tasks for all the TODOs that you created.

MarcellinoCO:
This commit addresses the change requests from @Elforama. Here is the newest screenshot:


I have solved the issues from my first comment, but have not done the Plant icon.
>     3. Need to update 'Plant' icon under user's phone number (the icon can't be exported in the design).

## Changes

**DepthButton changes**:
- Added `contentAlignment` parameter. This parameter is used to position the contents alignment inside the button.
- Changed the `DepthSurface` modifier's `fillMaxSize()` into `matchParentSize()`. This allows the `DepthSurface` to match its size to the contents of the button, and the use of `wrapContentSize()` in the modifier parameter.
- Added documentation to the component.

**UserButton changes**:
- Removed explicit button height, and used `wrapContentHeight()` instead.
- Decoupled the `UserSelectState`, replacing it with a boolean instead.

Elforama:
Great fixes, this is looking good.

Elforama (approved):

PR #631, opened 2021-05-19
Added the TopBarTitle to the adequate screens

brntsw:
I have added the component TopBarTitle on a separate Text.kt file (on view folder) and added it to Dashboard, Language Select and Sign up screens.

The result can be seen below:

Screen Recording 2021-05-19 at 19 44 43


Elforama (approved):
Great work!

Issue #614, opened 2021-04-18
Update Wallet Select Screen Header UX design


Elforama:
Update WalletSelectScreen.kt to header to include a selfie image in the top left corner to match the design found here: https://www.sketch.com/s/a0bbeac0-767a-4873-beef-c6c7ffd5826a/a/dljjlV9

This image should be of the currently logged in user. 
Clicking on the image should navigate to the SignupFlow.

PR #634, opened 2021-06-02
#614 - Update WalletSelectScreen to include selfie image in the top left corner

brntsw:
Implemented the selfie on the top left corner of the Wallet selection screen. Also added the user's image on that selfie and also on the user selection screen.

A demo of how that's behaving:

https://user-images.githubusercontent.com/7141687/120409660-c5868380-c327-11eb-9bfe-8a36b458eeb2.mov

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -304,6 +304,6 @@ afterEvaluate {
     // ktlint git precommit hook is insalled to .git folder in project root when running app in
     // Intellij by the developer or when triggering gradle build. This ensures ktlint format is
     // executed before code is committed.
-//    tasks["preBuild"].dependsOn parent.tasks.findByName("addKtlintFormatGitPreCommitHook")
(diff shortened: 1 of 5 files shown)


Elforama (changes requested):
This looks great! Just wondering if we still need the Box around the localimage. It looks like all those modifiers can be moved to the image instead.

brntsw:
I have removed the Box, it was really just a placeholder for the Image, I didn't realize that.
Also, I added the ContentScale parameter to LocalImage, as it was necessary to me to have that and I thought that would be unnecessary to create another Composable for that reason.

The behavior keeps the same:

https://user-images.githubusercontent.com/7141687/120472745-a6fea780-c37c-11eb-9adc-62d06a8742f1.mov

Elforama (approved):
Thanks for the quick update, great work!

Issue #635, opened 2021-06-05
Refresh UserSelect screen


Elforama:
When a new user is added and then you go back to the UserSelect screen, that new user doesn't show up.

The UserSelect screen should always show all users.

https://greenstand.slack.com/files/U01VCVBS7HQ/F023YFLL22V/screen_recording_2021-05-31_at_22.21.01.mov

PR #638, opened 2021-06-17
#635 Fixed the issue related to not refreshing the newly added user

brntsw:
The issue related to not adding the newly added user to the User select screen was fixed.

https://user-images.githubusercontent.com/7141687/122314814-3f556a00-ceef-11eb-9275-4e5415d55c42.mov


Elforama (changes requested):
Although this works, I think there is a better way to do this. The compose docs say a composable function can be run x amount of times, which means it's possible we can call the viewmodel x amount of times redundantly. 

There might be a way to achieve this with some kind of 'side effect' (https://developer.android.com/jetpack/compose/side-effects) Let me know what you think.

Elforama:
Solution that fixes issues here: #680

PR #639, opened 2021-06-25
Updated Language button to use Depth button

EngrTaofeek:
Hi Jonathan did this a long time ago but had issues making pull request directly from android studio. So I just made the edit directly on github since I was sure it was working already


Elforama:
@EngrTaofeek Can you please update this to pass tests?

PR #649, opened 2021-09-02
Capture flow completed

EngrTaofeek:
I have completed the capture flow as indicated in the issue assigned to me. The items now saves to the database and the back button takes the user to the dashboard.

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/camera/Camera.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/camera/Camera.kt
@@ -75,6 +75,7 @@ fun Camera(
                         )
                         .build()
 
+                    cameraProvider.unbindAll()
                     cameraProvider.bindToLifecycle(
                         lifecycleOwner, cameraSelector, preview, imageCapture)
 
--- a/app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureScreen.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureScreen.kt
@@ -48,7 +48,7 @@ fun TreeCaptureScreen(
             ActionBar(
                 leftAction = {
                     ArrowButton(isLeft = true) {
-                        navController.popBackStack()
+                        navController.navigate(NavRoute.Dashboard.route)
                     }
                 },
                 centerAction = {
--- a/app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureScreen.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureScreen.kt
@@ -47,9 +47,12 @@ fun TreeCaptureScreen(
         bottomBar = {
             ActionBar(
                 leftAction = {
-                    ArrowButton(isLeft = true) {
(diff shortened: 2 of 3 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/usecases/CreateTreeUseCase.kt
```diff
@@ -34,10 +34,11 @@ class CreateTreeUseCase(
...
         )
         analytics.treePlanted()
-        val attributeEntitites = params.treeCaptureAttributes().map {
-            TreeAttributeEntity(it.key, it.value, -1)
-        }.toList()
-        Timber.d("Inserting TreeCapture entity $entity")
+        val attributeEntitites: List<TreeAttributeEntity>? = params.treeCaptureAttributes().entries.map {
+            TreeAttributeEntity("${it.key}", "${it.value}" , -1)
+        }
+
```
Elforama, line 40: What was this part changed for?

app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureScreen.kt
```diff
@@ -48,7 +48,7 @@ fun TreeCaptureScreen(
             ActionBar(
                 leftAction = {
                     ArrowButton(isLeft = true) {
-                        navController.popBackStack()
+                        navController.navigate(NavRoute.Dashboard.route)
```
Elforama, line 51: This navigates back to the dashboard, but if I press 'back' on the dashboard I go to the camera screen again.

app/src/main/java/org/greenstand/android/TreeTracker/camera/Camera.kt
```diff
@@ -75,6 +75,7 @@ fun Camera(
                         )
                         .build()
 
+                    cameraProvider.unbindAll()
```
Elforama, line 78: This seems like it does a reset right before we load the camera. Would it work better if we unbindAll right after taking the picture?

Elforama (changes requested):
Overall, it's looking good. Just some things to address.

app/src/main/java/org/greenstand/android/TreeTracker/usecases/CreateTreeUseCase.kt
EngrTaofeek: It was actually not necessary, I have replaced it with the existing one.

app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureScreen.kt
EngrTaofeek: Okay fixed this in a new commit.

app/src/main/java/org/greenstand/android/TreeTracker/camera/Camera.kt
EngrTaofeek: I was getting an error that I was trying to bind too many usecases. So it wasn't able to take/save the picture and this fixed it.

Elforama (approved):
Great work, this looks good!

PR #656, opened 2021-09-27
Implemented WalletScreen to match current design

EngrTaofeek:
Implemented WalletScreen to match current design using two dummy images which would be updated later. Also hid the organization button. Was having issues navigating from the Clicking header profile image to user select sceen, it kept behaving strangely would appreciate any tips concerning hat while I try to find a fix around it. My proposed fix is to modify the LocalImage view to take an OnClick parameter.

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/walletselect/WalletSelectScreen.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/walletselect/WalletSelectScreen.kt
@@ -71,8 +71,9 @@ fun WalletSelectScreen(
                         isLeft = false,
                         isEnabled = state.selectedUser != null
                     ) {
-                        state.currentUser?.photoPath?.let {
-                            navController.navigate(NavRoute.TreeCapture.create(it))
+                        state.currentUser?.let { user ->
+                            viewModel.startSession(user)
+                            navController.navigate(NavRoute.TreeCapture.create(user.photoPath))
                         }
                     }
                 },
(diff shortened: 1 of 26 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/walletselect/WalletSelectScreen.kt
```diff
@@ -77,6 +80,26 @@ fun WalletSelectScreen(
...
+                                .size(height = 70.dp, width = 70.dp),
+                        ) {
+                            Text(
+                                buildAnnotatedString {
+                                    withStyle(style = SpanStyle(color = Color.Black,fontWeight = FontWeight.Bold, fontSize = 50.sp)) {
+                                        append("+")
+                                    }
+
+                                     }
+                            )
```
Elforama, line 99: Why build an annotated string? A regular string can work, the text object takes a style.

app/src/main/java/org/greenstand/android/TreeTracker/walletselect/WalletSelectScreen.kt
```diff
@@ -111,16 +133,37 @@ fun WalletSelectScreen(
...
+                        modifier = Modifier
+                            .height(150.dp)
+                            .width(150.dp)
+                            .padding(20.dp,10.dp,10.dp,30.dp,)
+
+
+
+            )
+
+
```
Elforama, line 157: place cleanup the spacing

app/src/main/java/org/greenstand/android/TreeTracker/walletselect/WalletSelectScreen.kt
```diff
@@ -111,16 +133,37 @@ fun WalletSelectScreen(
...
+            WalletUserButton(
+                user = user,
+                isSelected = isSelected,
+                AppButtonColors.Default,
+                AppColors.Green,
+                onClick = { onClick(user.id) },
+
```
Elforama, line 164: Please remove extra spacing

Elforama (changes requested):

Issue #661, opened 2021-10-12
Fix backstack navigation for profile icon


Elforama:
Repro:
- Click user icon
- Land on wallet selection screen
- Go back using system navigation
- BUG: user goes back to screen user select screen instead of dashboard
-- FIX: User should go to dashboard when exiting user selection

PR #665, opened 2021-10-22
NOTE FEATURE

EngrTaofeek:
Added note feature to tree capturing and fixed back navigation for and Fix backstack navigation for profile icon #661


Elforama:
Please make sure to test the app before creating a PR.

I tested the build a bit further. When I capture a tree image and click the check box to accept it, it then goes to the selfie screen. If I then press the selfie button, the app crashes.

PR #674, opened 2021-11-01
Fixed Bug that prevented user from adding creditials(email and phone) to the database.

EngrTaofeek:
The bug was as a result of having different userstate instances in the livedata from two composes. Other possible solutions could have been creating a singleton of the data class which is generally a bad practice. This was the reason for creating a new SignUpScreen compose file for both screens so they can always use the same instance of viewmodel and UserState


app/src/main/java/org/greenstand/android/TreeTracker/signup/SignupViewModel.kt
```diff
@@ -10,10 +10,17 @@ import org.greenstand.android.TreeTracker.utilities.Validation
...
     val organization: String? = null,
     val photoPath: String? = null,
     val credential: Credential = Credential.Email(),
-)
+   )
+data class CredentialState(
+    val nameEntryStage: Boolean = false,
+    val isEmailValid: Boolean = false,
+    val isPhoneValid: Boolean = false,
+    )
```
Elforama, line 23: Each viewmodel should have a single state object. You can move these 3 fields into the SignUpState object.

app/src/main/java/org/greenstand/android/TreeTracker/models/user/User.kt
```diff
@@ -5,6 +5,8 @@ data class User(
     val wallet: String,
     val firstName: String,
     val lastName: String?,
+    val email: String?,
+    val phone: String?,
```
Elforama, line 9: The email/phone is saved as a users "wallet" ID, so whichever the user decides to use should just be saved under "wallet"

app/src/main/java/org/greenstand/android/TreeTracker/signup/CredentialEntryView.kt
```diff
@@ -42,10 +42,11 @@ import org.greenstand.android.TreeTracker.view.LanguageButton
 import org.greenstand.android.TreeTracker.view.TopBarTitle
 
 @Composable
-fun SignupFlow(
+fun CredentialEntryView(
     viewModel: SignupViewModel = viewModel(factory = LocalViewModelFactory.current),
 ) {
     val state by viewModel.state.observeAsState(SignUpState())
+    val credentialState by viewModel.credentialState.observeAsState(CredentialState())
```
Elforama, line 49: One state per screen.

app/src/main/java/org/greenstand/android/TreeTracker/signup/NameEntryView.kt
```diff
@@ -117,7 +120,9 @@ fun NameEntryView(
                 ),
                 keyboardActions = KeyboardActions(
                     onGo = {
-                        cameraLauncher.launch(true)
+                        if (state.name != null && state.organization != null) {
+                            cameraLauncher.launch(true)
+                        }
```
Elforama, line 125: -nit-

We can try to remove all business logic from this view by making this if statement logic into an extension function on the state object. The extension function can contain that boolean logic.

if (state.canGoToNextScreen) {
   // go to next screen
}

What are your thoughts?

app/src/main/java/org/greenstand/android/TreeTracker/signup/SignupViewModel.kt
```diff
@@ -59,16 +68,22 @@ class SignupViewModel(private val users: Users) : ViewModel() {
...
-        _state.value = _state.value?.copy(credential = Credential.Phone().apply { text = phone })
+        _state.value = _state.value?.copy(phone = phone)
+        _credentialState.value = _credentialState.value?.copy(isPhoneValid = Validation.isValidPhoneNumber(phone))
+
     }
 
     fun updateCredentialType(updatedCredential: Credential) {
         _state.value = _state.value?.copy(credential = updatedCredential)
     }
+    fun updateSignUpState(state: Boolean){
```
Elforama, line 84: Please leave space between functions.

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -354,9 +354,10 @@ fun DepthSurfaceCircle(
 @Composable
 fun OrangeAddButton(
     modifier: Modifier,
-) {
+    onClick: () -> Unit,
+    ) {
```
Elforama, line 358: Please keep formatting as it was.

app/src/main/java/org/greenstand/android/TreeTracker/view/SelectableItemDetail.kt
```diff
@@ -84,48 +84,8 @@ fun UserButton(
...
-}
-@Composable
-fun WalletUserButton(
-    user: User,
-    isSelected: Boolean,
-    buttonColors: DepthButtonColors,
-    selectedColor: Color,
-    onClick: () -> Unit
-) {
+    val credential: String? = if (user.email.isNullOrEmpty()) user.phone else user.email
```
Elforama, line 87: This should be the users "wallet". Their wallet is the ID that the account is associated with. Atm this is an email/phone but in the future it could be something else such as "@myWalletName".

app/src/main/java/org/greenstand/android/TreeTracker/signup/SignupViewModel.kt
```diff
@@ -59,16 +68,22 @@ class SignupViewModel(private val users: Users) : ViewModel() {
 
     fun updateEmail(email: String) {
         // TODO validate data and show errors if needed, after click
-        _state.value = _state.value?.copy(credential = Credential.Email().apply { text = email })
+        _state.value = _state.value?.copy(email = email)
+        _credentialState.value = _credentialState.value?.copy(isEmailValid = Validation.isEmailValid(email))
```
Elforama, line 72: Updating two states like this will make the view update twice. This is one of the reasons why we should use 1 state per screen.

Elforama (changes requested):
This is looking good, just a few minor fixes before it's merged.

Please keep in mind the formatting of the code. Unified formatting makes things easier to read.

app/src/main/java/org/greenstand/android/TreeTracker/models/user/User.kt
EngrTaofeek: There is currently no wallet ID in the planterInfo Entity or should I modify the columns of the planter info table to suit this.

app/src/main/java/org/greenstand/android/TreeTracker/signup/NameEntryView.kt
EngrTaofeek: Yes this would be much better but I would suspend this for now. Gwen made some note about validations approach so the implementation would change soon.

app/src/main/java/org/greenstand/android/TreeTracker/view/SelectableItemDetail.kt
EngrTaofeek: This is subject to your response in https://github.com/Greenstand/treetracker-android/pull/674#discussion_r748724697

app/src/main/java/org/greenstand/android/TreeTracker/models/user/User.kt
Elforama: The field `identifier` in the PlanterInfoEntity is what we'll use as the wallet ID.

app/src/main/java/org/greenstand/android/TreeTracker/signup/NameEntryView.kt
Elforama: We can put it in another PR if you'd like. But it should be done. The 2.0 milestone for the app will have the functionality of the old one, which validates email and phone number.

Once we move on to 2.1+ we'll start adding in new features such as org flavors, messages screen, etc..

PR #682, opened 2021-11-14
Fix signup flow

Elforama:


EngrTaofeek (commented):
I don't seem to be able to identify any possible bug or issues here previously but this is more of refactoring and definitely much concise and better than what was in here previously.

PR #685, opened 2021-11-20
Permission management to prompt permission check and request permissi…

EngrTaofeek:
Permission management to prompt permission check and request permissions at run time from any where in the app.
REASONS FOR FILE modifications 
1 Compile and target sdk version had to be updated for the accompanist library.
2. Manifest was edited as android:exported had to be declared explicitly in versions 31 and above
3.  Also annotating compose functions and nav with @ExperimentalPermissionsApi
was also required.

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/permission/Permissions.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/permission/Permissions.kt
@@ -4,6 +4,7 @@ import android.Manifest
 import android.app.Activity
 import android.content.Context
 import android.content.Intent
+import android.net.Uri
 import androidx.compose.foundation.layout.*
 import androidx.compose.material.*
 import androidx.compose.runtime.*
@@ -18,17 +19,25 @@ import com.google.accompanist.permissions.rememberMultiplePermissionsState
 import android.provider.Settings
 import androidx.compose.material.Text
 import androidx.compose.runtime.Composable
+import androidx.compose.runtime.livedata.observeAsState
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.res.stringResource
 import androidx.core.content.ContextCompat.startActivity
+import androidx.lifecycle.viewmodel.compose.viewModel
+import androidx.navigation.NavHostController
 import org.greenstand.android.TreeTracker.R
+import org.greenstand.android.TreeTracker.permission.PermissionItemsState
+import org.greenstand.android.TreeTracker.permission.PermissionViewModel
 import org.greenstand.android.TreeTracker.root.LocalNavHostController
+import org.greenstand.android.TreeTracker.root.LocalViewModelFactory
 
 @ExperimentalPermissionsApi
 @Composable
-fun PermissionRequest() {
-
+fun PermissionRequest(
+    viewModel: PermissionViewModel = viewModel(factory = LocalViewModelFactory.current)
+) {
     val navController = LocalNavHostController.current
+    val state by viewModel.state.observeAsState(PermissionItemsState())
     val permissionsState = rememberMultiplePermissionsState(
         permissions = listOf(
             Manifest.permission.ACCESS_COARSE_LOCATION,
@@ -51,15 +60,15 @@ fun PermissionRequest() {
             }
         }
     )
+
     permissionsState.permissions.forEach { perm ->
         when (perm.permission) {
             Manifest.permission.CAMERA -> {
                 when {
                     perm.hasPermission -> {
-                        //Check if location is enabled from the LocationUpdateManager before calling enableLocation
                         return
                     }
-                    (perm.shouldShowRationale || !perm.permissionRequested) -> {
+                    perm.shouldShowRationale -> {
                         AlertDialog(
                             onDismissRequest = { navController.popBackStack() },
                             title = {
@@ -67,7 +76,7 @@ fun PermissionRequest() {
                             },
                             text = {
                                 Text(
-                                    text = stringResource(R.string.accept_location_permission_message),
+                                    text = stringResource(R.string.accept_camera_permission_message),
                                     color = Color.Green
                                 )
                             },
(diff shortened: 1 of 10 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/permission/Permissions.kt
```diff
@@ -0,0 +1,174 @@
...
+                }
+            }
+            Manifest.permission.ACCESS_COARSE_LOCATION -> {
+                when {
+                    perm.hasPermission -> {
+                        // TODO Check if location is enabled from the LocationUpdateManager before calling enableLocation
+                        return
+                    }
+                    (perm.shouldShowRationale || !perm.permissionRequested) -> {
+                        AlertDialog(
```
Elforama, line 134: To clean things up, let's move this location permission dialog to it's own function.

app/src/main/java/org/greenstand/android/TreeTracker/permission/Permissions.kt
```diff
@@ -0,0 +1,174 @@
...
+                    }
+                }
+            }
+        }
+    }
+}
+
+fun isLocationEnabledCheck() {
+
+}
```
Elforama, line 167: This is not needed.

app/src/main/java/org/greenstand/android/TreeTracker/permission/Permissions.kt
```diff
@@ -0,0 +1,174 @@
...
+
+@Composable
+fun enableLocation() {
+    val activity: Context = LocalContext.current as Activity
+    val intent = Intent(Settings.ACTION_LOCATION_SOURCE_SETTINGS);
+    startActivity(activity, intent, null)
+}
```
Elforama, line 174: This is not used anywhere, but it should be used to help a user who denied the location permission to re-enable it in settings.

app/src/main/java/org/greenstand/android/TreeTracker/permission/Permissions.kt
```diff
@@ -0,0 +1,174 @@
...
+                                    text = stringResource(R.string.accept_location_permission_message),
+                                    color = Color.Green
+                                )
+                            },
+                            buttons = {
+                                Button(
+                                    modifier = Modifier.wrapContentSize(),
+                                    onClick = {
+                                        perm.launchPermissionRequest()
+                                        navController.popBackStack()
```
Elforama, line 150: Why does this exit the screen? Seems to make more sense to request permission on camera screen like it initially did.

app/src/main/java/org/greenstand/android/TreeTracker/permission/Permissions.kt
```diff
@@ -0,0 +1,174 @@
...
+
+                }
+            }
+            Manifest.permission.ACCESS_COARSE_LOCATION -> {
+                when {
+                    perm.hasPermission -> {
+                        // TODO Check if location is enabled from the LocationUpdateManager before calling enableLocation
+                        return
+                    }
+                    (perm.shouldShowRationale || !perm.permissionRequested) -> {
```
Elforama, line 133: Looks like we handle the first time we'd show the rationale. However, if they deny the permission, they are stuck with it disabled. The dialog will says "accept permission" but nothing will happen.

To fix this I think we need to check for 'perm.permissionRequested'. If it's true, then we need to send them to the settings screen to enable it.

This applies to all the permissions we need.

Elforama (changes requested):
Looking good. Just a few things to fix.

make sure we send them to settings screen if they can't see dialog anymore.
Cleanup the dialogs to be their own functions.

app/src/main/java/org/greenstand/android/TreeTracker/permission/Permissions.kt
EngrTaofeek: This is actually needed but I need your help here. This function is to check if location/gps is enabled  which should return a boolean valueand it is only called after confirming permission has been accepted. My challenge is accessing fun isLocationEnabled() in the Locations.kt file(line 106) to return the value

app/src/main/java/org/greenstand/android/TreeTracker/permission/Permissions.kt
EngrTaofeek: This would be used after isLocationEnabled() is called so it prompts the user to turn on the gps/location if it was previously turned off.

app/src/main/java/org/greenstand/android/TreeTracker/permission/Permissions.kt
EngrTaofeek: It doesn't actually exit the screen but reloads the camera screen as the camera was initially called without permission and would show blank.

app/src/main/java/org/greenstand/android/TreeTracker/permission/Permissions.kt
EngrTaofeek: Yes that makes sense since both fine and coarse permissions use the same dialog.

app/src/main/java/org/greenstand/android/TreeTracker/permission/Permissions.kt
Elforama: The isLocationEnabled method in the LocationUpdateManager is in the "model" layer of the app. There is the view/viewmodel/model layers. The model should only talk to the viewmodel layer. That means you can add the LocationUpdateManager to a viewmodel, then expose that locationEnabled through a state object.

This permission method doesn't have a viewmodel, but in this case it could make life easier if we make a viewmodel for this, since it's used in multiple screens. To do that, follow the convention the other viewmodels use.

EngrTaofeek: Great suggestion, I would try this out.

app/src/main/java/org/greenstand/android/TreeTracker/permission/Permissions.kt
EngrTaofeek: I think I get this now, I would make a fix to this.

EngrTaofeek:
Updated this PR with another branch moving all changes to this new PR https://github.com/Greenstand/treetracker-android/pull/693

Elforama:
New PR made for this. Closing.

PR #690, opened 2021-11-25
fix: profile image bug for wallets

thePlebDev:
`state.selectedUser` would create a bug where the profile image would change with every click on a different wallet . `state.currentUser` means the profile image is constant with the current signed in user. Which is the desired effect.


Elforama (approved):
Thanks for the fix, looks great!

PR #692, opened 2021-11-26
replaced placeholder images and updated dashboard screen

EngrTaofeek:
Replaced placeholder images and updated dashboard screen.


Elforama (approved):
This looks great, thanks! Please make sure that it passes the tests, and I'll have it merged.

Elforama (changes requested):
The reason it doesn't compile is due to the icons being in the greenstand flavor. You need to add them to the 'main' res folder.

EngrTaofeek:
> The reason it doesn't compile is due to the icons being in the greenstand flavor. You need to add them to the 'main' res folder.

This was probably the same reason why my previous PR failed, I would fix them both now. Thanks for pointing out. I was already reading the failed log and trying to make some sense from it.

Elforama (changes requested):
Please remove the images from the greenstand files.

The way this works is that the app will use whatever is in the main folder. But if something is provided in the greenstand folder it will use it instead. Since we only want one copy of the icons for all builds, we should only have them in the main folder.

EngrTaofeek:
> Please remove the images from the greenstand files.
> 
> The way this works is that the app will use whatever is in the main folder. But if something is provided in the greenstand folder it will use it instead. Since we only want one copy of the icons for all builds, we should only have them in the main folder.

Done

Elforama (changes requested):
Please double check your PRs. The images are still in the Greenstand flavor.

EngrTaofeek:
> Please double check your PRs. The images are still in the Greenstand flavor.

The images were deleted in a previous commit but merging remote added them now. I have fixed it in a new commit https://github.com/Greenstand/treetracker-android/pull/692/commits/c36663cdb3afc10e9ca6202728625e4ec9e7fa93.

PR #693, opened 2021-11-29
Fixed permission management

EngrTaofeek:
Updated PR https://github.com/Greenstand/treetracker-android/pull/685
Created another branch and applied all changes in the previous PR above


app/src/main/java/org/greenstand/android/TreeTracker/permissions/Permissions.kt
```diff
@@ -0,0 +1,211 @@
...
+
+@Composable
+fun enableLocation() {
+    val activity: Context = LocalContext.current as Activity
+    val intent = Intent(Settings.ACTION_LOCATION_SOURCE_SETTINGS);
+    startActivity(activity, intent, null)
+}
+
+@Composable
+fun PermissionDeniedPermanentlyDialog(navController: NavHostController,viewModel: PermissionViewModel = viewModel(factory = LocalViewModelFactory.current)) {
```
Elforama, line 179: This viewmodel is never used

app/src/main/java/org/greenstand/android/TreeTracker/permissions/Permissions.kt
```diff
@@ -0,0 +1,211 @@
...
+    val activity: Context = LocalContext.current as Activity
+    val intent = Intent(
+        Settings.ACTION_APPLICATION_DETAILS_SETTINGS,
+        Uri.fromParts("package", "org.greenstand.android.TreeTracker.debug", null)
+    )
+    intent.addFlags(Intent.FLAG_ACTIVITY_NEW_TASK)
```
Elforama, line 185: We can create this intent once the user clicks the button. Otherwise we create it for no reason.

app/src/main/java/org/greenstand/android/TreeTracker/permissions/Permissions.kt
```diff
@@ -0,0 +1,211 @@
...
+                    }
+                    else -> {
+                            PermissionDeniedPermanentlyDialog(navController)
+                    }
+                }
+            }
+            Manifest.permission.ACCESS_COARSE_LOCATION -> {
+                when {
+                    perm.hasPermission -> {
+                        if(viewModel.isLocationEnabled() == false){
```
Elforama, line 123: There is a lint warning here. No need to check equality. Please update.

app/src/main/java/org/greenstand/android/TreeTracker/permissions/Permissions.kt
```diff
@@ -0,0 +1,211 @@
...
+                            enableLocation()
+                        }
+                        return
+                    }
+                    perm.shouldShowRationale -> {
+                        LocationRationaleDialog(navController = navController, perm = perm)
+                    }
+                    else -> {
+                            PermissionDeniedPermanentlyDialog(navController)
```
Elforama, line 132: Please keep spacing/indentation consistent.

app/src/main/java/org/greenstand/android/TreeTracker/permissions/Permissions.kt
```diff
@@ -0,0 +1,211 @@
...
+                        if(state.isLocationEnabled == false){
+                            enableLocation()
+                        }
+                        return
+                    }
+                    perm.shouldShowRationale -> {
+                        LocationRationaleDialog(navController = navController, perm = perm)
+                    }
+                    else -> {
+                            PermissionDeniedPermanentlyDialog(navController)
```
Elforama, line 116: Spacing/indentation

app/src/main/java/org/greenstand/android/TreeTracker/permissions/PermissionViewModel.kt
```diff
@@ -0,0 +1,26 @@
+package org.greenstand.android.TreeTracker.permission
+
+import android.location.LocationManager
+import androidx.lifecycle.LiveData
+import androidx.lifecycle.MutableLiveData
+import androidx.lifecycle.ViewModel
+
+data class PermissionItemsState(
+    val openAppSettings: Boolean = true,
+    val isLocationEnabled: Boolean? = null
```
Elforama, line 10: I think we can set this to false by default.

app/src/main/java/org/greenstand/android/TreeTracker/permissions/PermissionViewModel.kt
```diff
@@ -0,0 +1,26 @@
...
+)
+class PermissionViewModel (
+    private val locationManager: LocationManager) : ViewModel(){
+
+    private val _state = MutableLiveData(PermissionItemsState(isLocationEnabled = isLocationEnabled()))
+    val state: LiveData<PermissionItemsState> = _state
+
+    fun isLocationEnabled(): Boolean {
+        return locationManager.isProviderEnabled(LocationManager.GPS_PROVIDER)
+    }
```
Elforama, line 20: The view should not call method on the viewmodel to get info, it should only rely on the state object. In this case, when the viewmodel is initialized, it can query the GPS status and update the state with the latest value. I'm assuming this was why you have the "isLocationEnabled" in the state object.

Looking at line 107 and 123 I can see the same logic done in two ways. One using state, the other using the viewmodel.

Elforama (changes requested):
This is looking good, there just needs to be a bit of cleanup and more usage of the state object.

app/src/main/java/org/greenstand/android/TreeTracker/permissions/PermissionViewModel.kt
EngrTaofeek: The parameter "openAppSettings" is no longer being used as a state object so I deleted it in a new commit.

app/src/main/java/org/greenstand/android/TreeTracker/permissions/Permissions.kt
EngrTaofeek: Right not an efficient way. This has been replaced with the state object.

app/src/main/java/org/greenstand/android/TreeTracker/permissions/Permissions.kt
EngrTaofeek: Changes applied.

app/src/main/java/org/greenstand/android/TreeTracker/permissions/Permissions.kt
EngrTaofeek: The portion of code using it previously has been deleted. I would delete this as well.

app/src/main/java/org/greenstand/android/TreeTracker/permissions/PermissionViewModel.kt
Elforama: I was refering to the isLocationEnabled. It's set to null, but can be set to false by default. This will allow you to avoid dealing with nullability. 
It's why you have 
if(state.isLocationEnabled == false)
instead of
if(state.isLocationEnabled)

app/src/main/java/org/greenstand/android/TreeTracker/permissions/Permissions.kt
```diff
@@ -0,0 +1,211 @@
...
+
+    permissionsState.permissions.forEach { perm ->
+        when (perm.permission) {
+            Manifest.permission.CAMERA -> {
+                when {
+                    perm.hasPermission -> {
+                        return
+                    }
+                    perm.shouldShowRationale -> {
+                        AlertDialog(
```
Elforama, line 72: Can you move this alert dialog to its own function.

app/src/main/java/org/greenstand/android/TreeTracker/permissions/Permissions.kt
```diff
@@ -0,0 +1,211 @@
...
+
+                }
+            }
+            Manifest.permission.ACCESS_FINE_LOCATION -> {
+                when {
+                    perm.hasPermission -> {
+                        if(state.isLocationEnabled == false){
+                            enableLocation()
+                        }
+                        return
```
Elforama, line 110: No need for return statement in your when statements.

app/src/main/java/org/greenstand/android/TreeTracker/permissions/Permissions.kt
```diff
@@ -0,0 +1,211 @@
...
+                    }
+
+                }
+            }
+            Manifest.permission.ACCESS_FINE_LOCATION -> {
+                when {
+                    perm.hasPermission -> {
+                        if(state.isLocationEnabled == false){
+                            enableLocation()
+                        }
```
Elforama, line 109: This can be
perm.hasPermission && state.isLocationEnabled -> enableLocation()

Elforama (changes requested):
Looking good! Just a few small changes and this is good to merge.

app/src/main/java/org/greenstand/android/TreeTracker/permissions/Permissions.kt
EngrTaofeek: if isLocationEnabled: Boolean? is set to false as default it always open the location settings screen as false remains the value before isLocationEnabled() function is run.

app/src/main/java/org/greenstand/android/TreeTracker/permissions/Permissions.kt
EngrTaofeek: If I refactor the code to this and location is enabled on the phone the else in the when block will run which shouldn't be so if permissions have been granted. So I think it is still in a way better like this.

app/src/main/res/values/strings.xml
```diff
@@ -79,4 +79,15 @@
     <string name="lat_std_dev_threshold" translatable="false">Latitude Standard Dev Threshold</string>
     <string name="add_note_to_tree">Add note to tree</string>
     <string name="note">NOTE</string>
+
+    <!--Strings for the permission management and rationale-->
+    <string name="accept_camera_permission_header">Accept Camera permission</string>
+    <string name="accept_camera_permission_message">Camera is needed for this app to allow it capture pictures, accept this permission to use this device to take pictures</string>
```
nationalkim, line 85: nit; I would maybe try to make this text more concise, something like "Camera is needed for Treetracker to capture pictures, please accept this permission"

app/src/main/res/values/strings.xml
```diff
@@ -79,4 +79,15 @@
...
+    <string name="accept_permission">Accept permission</string>
+    <string name="accept_location_permission_header">Accept Location permission</string>
+    <string name="accept_location_permission_message">Location is needed for this app to allow it track trees, Accept this permission to use the tree tracking feature</string>
```
nationalkim, line 88: ditto above

app/src/main/res/values/strings.xml
```diff
@@ -79,4 +79,15 @@
...
+    <string name="open_settings_header">Accept Permissions</string>
+    <string name="open_settings_message">You have denied this permission previously and would have to accept the permissions from the app settings. Click open settings below to accept the permissions required.</string>
```
nationalkim, line 90: ditto above; maybe something like "Permission currently denied. Click "Open Settings" below to accept the required permissions." ?

nationalkim (commented):
nice!

Elforama (approved):
Thanks for addressing everything. Once this is rebased it can be merged.

PR #697, opened 2021-12-06
User select and wallet select button

EngrTaofeek:
Updated user and wallet selection buttons to reflect number of trees planted by each planter. Also updated the placeholder image.


app/src/main/java/org/greenstand/android/TreeTracker/view/SelectableItemDetail.kt
```diff
@@ -102,13 +105,15 @@ fun UserButton(
...
         ) {
-            Box( // TODO: Change into 'Plant' icon.
+            Image(
+                painter = painterResource(id = R.drawable.white_leaf),
+                contentDescription = "",
                 modifier = Modifier
-                    .size(width = 20.dp, height = 22.dp)
-                    .background(AppColors.LightGray)
+                    .align(Alignment.CenterVertically)
+                    .size(width = 20.dp,height = 22.dp)
```
Elforama, line 113: Please be aware of formatting. You removed a space before the 'height' param.

Issue #698, opened 2021-12-06
Weak GPS signal while tracking trees


EngrTaofeek:
If a user attempts to use the app to capture trees where there is weak GPS signal, the app crashes. Common scenario is using the tracking indoors where gps is generally weak. The app crashes after showing convergence timed out in the logs.

From the Locations.kt file, handle situation where convergence times out and show user message or dialog showing possible reasons for failing to track.
Also prevent users from saving images of trees without locations.

arunbakt:
@EngrTaofeek , I tried to modify the configuration for convergence timeout and convergence thresholds and triggered the converge timeout in the app. But the app didn't crash while taking the tree pictures in my Google pixel as well as samsung galaxy.
Seems like this isn't a convergence triggered crash, must be something else. Did you observe it in a specific phone? Is there more information on this? Thank you!

arunbakt:
If you are not aware of the ability to mofidy the config values, you can get to it by double tapping on the build link in the about page of the app. A password prompt will pop up, (xxxx pinged in slack) is the password to change its values.

EngrTaofeek:
Hi @arunbakt I made this PR https://github.com/Greenstand/treetracker-android/pull/699 to fix this issue. Kindly check it out and the commit message.

arunbakt:
Hi @EngrTaofeek , I see the code base I had was older one compared to what exists in master(main) branch. I also see Jonathan has reviewed and requested changes to your PR. So I will let him merge the changes once you address those changes requested.

PR #699, opened 2021-12-07
Weak gps signal while tracking trees

EngrTaofeek:
PR to address https://github.com/Greenstand/treetracker-android/issues/698
This prevents app from crashing when gps signal is poor(lastConvergenceWithinRange and currentConvergence are null) and convergence times out. It also informs user of poor GPS connection and possible recommendations to avoid it. Additionally,it prevents going to Image review screen(saving tree) if location data was not captured.


Elforama (changes requested):
If we cannot get the GPS signal, then we should continue on without it.

app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureScreen.kt
```diff
@@ -33,7 +27,8 @@ import org.greenstand.android.TreeTracker.view.*
 fun TreeCaptureScreen(
     profilePicUrl: String,
 ) {
-    val viewModel: TreeCaptureViewModel = viewModel(factory = TreeCaptureViewModelFactory(profilePicUrl))
+    val viewModel: TreeCaptureViewModel =
+        viewModel(factory = TreeCaptureViewModelFactory(profilePicUrl))
```
Elforama, line 31: Please look over your PRs after you make them, to make sure there are no formatting issues like this. It makes it harder to review code when there are many little formatting changes that distract from what you actually did.

app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureViewModel.kt
```diff
@@ -26,7 +27,7 @@ class TreeCaptureViewModel(
 
     suspend fun captureLocation() {
         _state.value = _state.value?.copy(isGettingLocation = true)
-        treeCapturer.pinLocation()
+        _state.value = _state.value?.copy(isLocationAvailable = treeCapturer.pinLocation())
```
Elforama, line 30: Please only update the state with a single copy method.

app/src/main/java/org/greenstand/android/TreeTracker/models/TreeCapturer.kt
```diff
@@ -16,15 +16,20 @@ class TreeCapturer(
...
 
         newTreeUuid = locationDataCapturer.generatedTreeUuid
-        convergence = locationDataCapturer.convergence()
-
-        locationDataCapturer.turnOffTreeCaptureMode()
+        return if (locationDataCapturer.lastConvergenceWithinRange != null || locationDataCapturer.currentConvergence != null) {
+            convergence = locationDataCapturer.convergence()
+            locationDataCapturer.turnOffTreeCaptureMode()
+            true
+        }else{
```
Elforama, line 29: Please check your formatting.

app/src/main/java/org/greenstand/android/TreeTracker/models/Locations.kt
```diff
@@ -143,8 +143,8 @@ class LocationDataCapturer(
     private var locationsDeque: Deque<Location> = LinkedList()
     var generatedTreeUuid: UUID? = null
         private set
-    private var lastConvergenceWithinRange: Convergence? = null
-    private var currentConvergence: Convergence? = null
+    var lastConvergenceWithinRange: Convergence? = null
+    var currentConvergence: Convergence? = null
```
Elforama, line 147: This class was working fine without these being exposed. Can we find a way to fix the issue without exposing internal values?

app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureViewModel.kt
EngrTaofeek: Not sure I get this. Do you mean each method should only update a single state object?

app/src/main/java/org/greenstand/android/TreeTracker/models/Locations.kt
EngrTaofeek: It needs to be exposed so as to be used in the TreeCapturer.kt file to check if locations were captured at all after tree tracking. I added a private setter property to it in a newer commit, so it can only be read and not modified from outside the Locations.kt file.

EngrTaofeek:
> If we cannot get the GPS signal, then we should continue on without it.

In that situation there won't be location attached to the image of the tree which makes tracking pointless. So I felt it was better this way as the user can be immediately informed that the location was not captured and possible reasons why. It also saves the admin a lot of time as I guess only tree images with valid location can be cleared for payment.
Beyond that, the app crashes as the Review screen opens with a null location. I didn't bother debugging that has the previous fix prevented it from opening if locations were null.

app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureViewModel.kt
Elforama: Each time the state object is updated, it refreshes the screen. In this case you update it once per field, which updates the screen two times here.

To avoid this you can update the _state.value once with a single copy method that updates all the required fields in it.

EngrTaofeek: Oh I get this now.

Elforama:
We need to allow the user to capture a tree even if GPS can't converge. As long as GPS is on, we should take a picture. The alternative is someone plants 100 trees and then can't take pictures of them.

EngrTaofeek:
> We need to allow the user to capture a tree even if GPS can't converge. As long as GPS is on, we should take a picture. The alternative is someone plants 100 trees and then can't take pictures of them.

I don't quite understand the requested changes. What should be done if the location doesn't converge and is null? Previously, what was done in this PR was to allow camera take pictures but prevent trees from saving and show dialog to the planter telling them location was not gotten so they retry.

Elforama:
There is a 15 second timeout on the convergence. If it does not converge within that timeout, we allow them to take a picture anyone. You can check out the other files that use the convergence logic to see how the old version of the app was doing this.

EngrTaofeek:
> There is a 15 second timeout on the convergence. If it does not converge within that timeout, we allow them to take a picture anyone. You can check out the other files that use the convergence logic to see how the old version of the app was doing this.

I checked the use case in the last app and downloaded the tree tracker on play store apparently it is still an issue. The weird crash also exists in the play store version it just isn't a common use case that's why it isn't so obvious. I think this would be best discussed over during the next android/mobile meeting.

app/src/main/java/org/greenstand/android/TreeTracker/models/TreeCapturer.kt
```diff
@@ -16,15 +16,20 @@ class TreeCapturer(
...
+    suspend fun pinLocation(): Boolean {
         locationDataCapturer.start()
         locationDataCapturer.turnOnTreeCaptureMode()
         locationDataCapturer.converge()
 
         newTreeUuid = locationDataCapturer.generatedTreeUuid
-        convergence = locationDataCapturer.convergence()
-
-        locationDataCapturer.turnOffTreeCaptureMode()
+        return if (locationDataCapturer.lastConvergenceWithinRange != null || locationDataCapturer.currentConvergence != null) {
```
Elforama, line 25: Put this logic inside the dataCapturer

Elforama (changes requested):

app/src/main/java/org/greenstand/android/TreeTracker/models/Locations.kt
```diff
@@ -214,6 +216,10 @@ class LocationDataCapturer(
         }
     }
 
+    fun isLocationCoordinateAvailable(): Boolean {
+        return (lastConvergenceWithinRange != null || currentConvergence != null)
```
Elforama, line 220: <nit> no need to parenthesis

app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureViewModel.kt
```diff
@@ -30,8 +31,7 @@ class TreeCaptureViewModel(
 
     suspend fun captureLocation() {
         _state.value = _state.value?.copy(isGettingLocation = true)
-        treeCapturer.pinLocation()
-        _state.value = _state.value?.copy(isGettingLocation = false)
+        _state.value = _state.value?.copy(isLocationAvailable = treeCapturer.pinLocation(), isGettingLocation = false)
```
Elforama, line 34: This breaks the UI. 

Before: Clicking capture button shows loading spinner
Now: Clicking capture button shows nothing

Elforama (changes requested):

PR #702, opened 2021-12-09
Create Messages API

Elforama:
Sets up basic networking layer for messages API
Implements /messages endpoint

The end goal here is to have a MessagesRepo that gets message objects from either the network, or a db.


EngrTaofeek (changes requested):
This is a great way to start the messaging feature.

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/network/MessagesApiService.kt
```diff
@@ -0,0 +1,12 @@
+package org.greenstand.android.TreeTracker.models.messages.network
+
+import org.greenstand.android.TreeTracker.models.messages.network.responses.MessagesResponse
+import retrofit2.http.GET
+import retrofit2.http.Query
+
+interface MessagesApiService {
+
+    @GET("messaging/message")
+    suspend fun getMessages(@Query("author_handle") wallet: String): MessagesResponse
```
EngrTaofeek, line 10: Oh great this explains why there was no need for request bodies.

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/network/responses/MessageResponse.kt
```diff
@@ -0,0 +1,17 @@
...
+    val from: String,
+    val to: String,
+    val subject: String,
+    val body: String,
+    @SerializedName("composed_at")
+    val composedAt: String,
+    @SerializedName("parent_message_id")
+    val parentMessageId: String?,
+    @SerializedName("video_link")
+    val videoLink: String?,
```
EngrTaofeek, line 15: The video link here is it a URL? Would it be viewed with YouTube or Exoplayer?

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/network/RetrofitBuilder.kt
```diff
@@ -0,0 +1,26 @@
...
+class RetrofitBuilder {
+
+    fun create(): Retrofit {
+        return Retrofit.Builder()
+            .client(
+                OkHttpClient.Builder()
+                .addInterceptor(HttpLoggingInterceptor()
+                    .also { it.setLevel(HttpLoggingInterceptor.Level.BODY) })
+                .build())
+            .baseUrl(BASE_ENDPOINT)
```
EngrTaofeek, line 17: I feel it would be better if we moved the base url to a separate data class in such a way we can have new additional endpoints all in the same class.

EngrTaofeek (commented):
This looks very good and explain in details what the messaging feature would consists of.

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/network/RetrofitBuilder.kt
Elforama: We can allow multiple endpoints later on.

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/network/responses/MessageResponse.kt
Elforama: With YouTube

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/network/MessagesApiService.kt
Elforama: What do you mean?

PR #710, opened 2021-12-14
Bug sign up isssue 701

### LINKED ISSUE #701: BUG: Can create more than one account with same email/phone
Currently a user can create more than one account using the same email/phone. We need to make sure that when entering an email that is already registered on the device, we warn the user and stop them from continuing.

EngrTaofeek:
Fixed bug that allows user to bypass credential validation by pressing Go on the keyboard which
 skips validation.

Fixed bug that allowed user to create  multiple accounts with same plant identifier.
Fixes #701


app/src/main/java/org/greenstand/android/TreeTracker/database/TreeTrackerDAO.kt
```diff
@@ -33,9 +33,15 @@ interface TreeTrackerDAO {
     @Query("SELECT * FROM planter_info WHERE _id = :id")
     suspend fun getPlanterInfoById(id: Long): PlanterInfoEntity?
 
+    @Query("SELECT * FROM planter_info WHERE planter_identifier = :identity")
+    suspend fun getPlanterInfoById(identity: String): PlanterInfoEntity?
```
Elforama, line 37: Can we rename this to getPlanterInfoByIdentifier? Otherwise it can get confusing with the other method of the same name.

app/src/main/java/org/greenstand/android/TreeTracker/database/TreeTrackerDAO.kt
```diff
@@ -33,9 +33,15 @@ interface TreeTrackerDAO {
...
+
     @Query("SELECT * FROM planter_info WHERE power_user = 1")
     suspend fun getPowerUser(): PlanterInfoEntity?
 
+    @Query("SELECT COUNT(*) FROM planter_info WHERE planter_identifier = :identifier LIMIT 1")
+    suspend fun checkIfPlanterExists(identifier: String): Int
```
Elforama, line 43: We don't need this method. We can use the other method you created, and check if it returns null or not.

app/src/main/java/org/greenstand/android/TreeTracker/models/user/Users.kt
```diff
@@ -87,6 +93,10 @@ class Users(
         }
     }
 
+    suspend fun checkPlanterExists(identifier: String): Boolean{
+        return dao.checkIfPlanterExists(identifier) >= 1
+    }
```
Elforama, line 98: No need for this, just check for a nonnull user from getUserWithIdentifier. There can only ever be one user per identifier/wallet

app/src/main/java/org/greenstand/android/TreeTracker/signup/CredentialEntryView.kt
```diff
@@ -105,11 +115,11 @@ fun CredentialEntryView(viewModel: SignupViewModel, state: SignUpState) {
                     placeholder = { Text(text = stringResource(id = R.string.email_placeholder), color = Color.White) },
                     keyboardOptions = KeyboardOptions(
                         keyboardType = KeyboardType.Email,
-                        imeAction = ImeAction.Go,
+                        imeAction = ImeAction.Done,
```
Elforama, line 118: Why did you change this? 'done' usually means you completed something, but in this case we go to the next input.

app/src/main/java/org/greenstand/android/TreeTracker/signup/CredentialEntryView.kt
```diff
@@ -175,6 +185,53 @@ fun <T : Credential> CredentialButton(
...
+                UserButton(
+                    user = it,
+                    isSelected = false,
+                    AppButtonColors.Default,
+                    Green
+                ) {
+                    navController.navigate(NavRoute.Dashboard.route) {
+                        popUpTo(NavRoute.Language.route) { inclusive = true }
+                        launchSingleTop = true
+                    }
```
Elforama, line 218: It looks like if we click on an existing user, we then instantly log them out. We should log them in, and go to the wallet screen.

Elforama (changes requested):
Looking good. Just a few things to fix.

app/src/main/java/org/greenstand/android/TreeTracker/database/TreeTrackerDAO.kt
EngrTaofeek: Okay I didn't know there was an existing method with same name

app/src/main/java/org/greenstand/android/TreeTracker/models/user/Users.kt
```diff
@@ -94,7 +94,7 @@ class Users(
     }
 
     suspend fun checkPlanterExists(identifier: String): Boolean{
```
Elforama, line 96: rename to doesUserExist() or something with "user" in the name.

app/src/main/java/org/greenstand/android/TreeTracker/signup/CredentialEntryView.kt
```diff
@@ -212,8 +212,8 @@ fun ExistingUserDialog(
                     AppButtonColors.Default,
                     Green
                 ) {
-                    navController.navigate(NavRoute.Dashboard.route) {
-                        popUpTo(NavRoute.Language.route) { inclusive = true }
+                    navController.navigate(NavRoute.WalletSelect.create(it.id)) {
+                        popUpTo(NavRoute.SignupFlow.route) { inclusive = true }
```
Elforama, line 216: Let's rename "it" to "user" so we know what it is easier.

Elforama (changes requested):

Elforama (approved):
Looks great! Thank you

PR #714, opened 2021-12-20
Update top bar text to image

### LINKED ISSUE #706: Need for a better top section UX
This top section needs to not break up Treetracker and the language selector could be in another line rather. Ideally we also show the greenstand logo here next to Treetracker.
Screenshot_20211210_100307_org greenstand android TreeTracker test

EngrTaofeek:
Replaced Treetracker text in Action bar to use image.
Replaced text in NameEntryView.kt action bar to use TopBarTitle() composable
This fixes #706


app/src/main/java/org/greenstand/android/TreeTracker/view/Text.kt
```diff
@@ -24,10 +28,12 @@ object TextStyles {
 
 @Composable
 fun BoxScope.TopBarTitle() {
-    Text(
-        "TREETRACKER",
-        modifier = Modifier.align(Alignment.Center),
```
Elforama, line 29: We want to keep the center alignment. You'll noticed the icon is off center when you look at it.

Elforama (changes requested):

Elforama (approved):

PR #715, opened 2021-12-20
Sync wrong success message

EngrTaofeek:
Snackbar previously showed Sync failed after syncing trees. This commit fixes it to show correct message.


Elforama (approved):
Looks good

PR #720, opened 2021-12-26
#696: enable credential validation when using keyboard navigation

### LINKED ISSUE #696: BUG: Email/Phone accepts invalid email/phone
Issue: Email/Phone screen accepts invalid email/phone if user moves to next screen via the keyboard instead of the arrow buttons.

Task: The keyboard should require validation like the green arrow button does.

nationalkim:
closes #696 
criteria:
- keyboard "go" button requires same validation as the arrow button
- if credential is not valid, show error message

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/signup/CredentialEntryView.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/signup/CredentialEntryView.kt
@@ -15,12 +15,15 @@ import androidx.compose.material.AlertDialog
 import androidx.compose.material.Button
 import androidx.compose.material.Scaffold
 import androidx.compose.material.Text
+import androidx.compose.material.rememberScaffoldState
 import androidx.compose.runtime.Composable
 import androidx.compose.runtime.remember
+import androidx.compose.runtime.rememberCoroutineScope
 import androidx.compose.ui.Alignment
 import androidx.compose.ui.Modifier
 import androidx.compose.ui.focus.FocusRequester
 import androidx.compose.ui.graphics.Color
+import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.platform.LocalFocusManager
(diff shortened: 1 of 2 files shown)


EngrTaofeek (changes requested):
This looks great Natalie. The CustomToast was used in the old version of the app. I don't think we would want to use that in the 2.0 version. You can replace it with a snackbar or ignore the error message until the design team comes up with a custom error dialog.

app/src/main/java/org/greenstand/android/TreeTracker/signup/CredentialEntryView.kt
```diff
@@ -109,7 +110,11 @@ fun CredentialEntryView(viewModel: SignupViewModel, state: SignUpState) {
                     ),
                     keyboardActions = KeyboardActions(
                         onGo = {
-                            viewModel.goToNameEntry()
+                            if (state.isEmailValid || state.isPhoneValid) {
+                                viewModel.goToNameEntry()
+                            } else {
+                                CustomToast.showToast("Please enter a valid email")
```
Elforama, line 116: Please use string resources for any text.

Elforama (changes requested):

app/src/main/java/org/greenstand/android/TreeTracker/signup/CredentialEntryView.kt
nationalkim: @Elforama thanks! can you comment at all on which is the preferred way to go about things (i.e. making a snackbar or leaving a todo) regarding @EngrTaofeek 's suggestion above?

Elforama: Let's use the snackbar.

ZavenArra:
This PR has been incorporated elsewhere.  Thanks for the great work!

PR #721, opened 2021-12-27
Add text message method to repo

Elforama:
Added method to allow viewmodels to fetch the latest text messages for a given wallet and those in the convo.

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/models/messages/Converters.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/models/messages/Converters.kt
@@ -12,7 +12,7 @@ fun QuestionResponse.toQuestion(): Question {
 
 fun MessageResponse.toMessage(): Message {
     return when {
-        survey.questions.isNotEmpty() && survey.questions.first().prompt != null ->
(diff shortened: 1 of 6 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/models/messages/Converters.kt
```diff
@@ -5,21 +5,24 @@ import org.greenstand.android.TreeTracker.models.messages.network.responses.Ques
...
 fun MessageResponse.toMessage(): Message {
     return when {
-        survey.questions != null -> return SurveyMessage(
-            from = from,
-            to = to,
-            composedAt = composedAt,
-            questions = survey.questions.map { it.toQuestion() },
-            answers = survey.answers!!
-        )
+        survey.questions.isNotEmpty() && survey.questions.first().prompt != null ->
```
EngrTaofeek, line 15: survey.questions.first().prompt could be "NULL PROMPT" as per changes in line 8 of the Converters.kt file and wrongfully satisfy this condition since it is not null.

EngrTaofeek (changes requested):
Every other thing looks great. The messaging feature is getting ready gradually and more complex, I think it would be nice if we can create a wiki to lay out the roadmap/implementation of this feature so others can follow.

PR #722, opened 2022-01-06
Image resolution size modification

### LINKED ISSUE #711: Image size too small for plant identification
Some time between July 2021 and Oct 2021 the size of all the images taken in Haiti dropped to 360 x 480 px. I also checked for Tanzania and all recent images I sampled are also 360 x 480 px. This size is too small to allow reliable seedling identification by human eye. Previous images were usually 960 x 720 px (or larger), which is a sufficiently large image for identification. Please reconsider the image size if species labeling is a goal.

EngrTaofeek:
Changed minimum resolution(close to HD version depending on device) to scale well and also added compression to the image to reduce storage file without affecting quality. After testing across four devices I noticed compression didn't affect quality of image until it reaches about 40% which is why I left it at 70%. This fixes #711


app/src/main/java/org/greenstand/android/TreeTracker/utilities/ImageUtils.kt
```diff
@@ -361,19 +361,11 @@ object ImageUtils {
...
-        if (imageHeight > imageWidth) {
-            requiredWidth = 600
-        }
 
         var sampleSize = ceil((imageWidth.toFloat() / requiredWidth.toFloat()).toDouble()).toInt()
 
-        // If the original image is smaller than required, don't sample
-        if (sampleSize < 1) {
-            sampleSize = 1
-        }
```
Elforama, line 375: Why was this removed? From what I understand, this should ensure a small image is not made smaller.

EngrTaofeek: If the sampleSize is greater than one, it scales down the image then when it is less than it scales up the image. This logic actually prevents from scaling beyond 800 pixels so it uses as it was captured.
The previous implementation was more focussed on not exceeding 800 pixels rather than setting a minimum resolution. Find attached this image from the android library itself explaining the sample size.
Screenshot 2022-01-07 at 10 10 56

Elforama: The documentation actually says if it's less than 1, it doesn't do any scaling. This means the code you removed shouldn't change anything, since it did that internally already. Looks good to me.

PR #746, opened 2022-01-10
Size of button backdrop #737

### LINKED ISSUE #737: Size of button backdrop and button graphic is wrong on camera
Correct spec:

image

Incorrect in application

image

EngrTaofeek:
Fixes #737  issue. Also refactored the camera capture button to make it more easily reusable Additionally, I made the depth for the Image capture button 10F instead of the default 20f value so as to replicate the design more closely.


EngrTaofeek:
this is what the capture button  looks like with depth of 10f 
W

EngrTaofeek:
While it looks like this with the default depth of 20f
WhatsApp Image 2022-01-10 at 16 38 24

EngrTaofeek:
You can check out the linked issue #737 for more details.

Elforama (changes requested):
Looks like there is a duplicated image capture button.

Issue #739, opened 2022-01-07
Name/Organization screen shows briefly after selfie image capture


ZavenArra:
Name/Organization screen shows briefly after selfie image capture.  We should just see the dashboard load immediately.

Can provide a video if this bug description is unclear.

Amritah9:
@ZavenArra and @EngrTaofeek - Please take a look at this issue to see if it can be closed? Looks like there were two closed PRs to fix this this issue. Or is there anything else needed on this?  
I've moved this issue out of 2.2 for now.

PR #752, opened 2022-01-16
Credential /Name screen flash after selfie

EngrTaofeek:
(Not ready to be merged)I am trying to make a progressbar show instead of the flashes of the screen as highlited in issue #739 . But I am facing an issue accesing the SignupViewModel.kt, it shows no compile error but as soon as ImageReviewScreen.kt is to come on it shows white and doesn't review.  See attached screenshot on error showed. 
Screenshot 2022-01-16 at 14 41 00
Any tip on this?


Elforama:
I'm not sure this is the correct approach. Now instead of the signups screen flashing, a loading screen will flash. Maybe we can move the logic that saves the user to the last screen so there is no need to go back.

EngrTaofeek:
> I'm not sure this is the correct approach. Now instead of the signups screen flashing, a loading screen will flash. Maybe we can move the logic that saves the user to the last screen so there is no need to go back.

that sounds better

ZavenArra:
Was a solution to this issue found and implemented?

Elforama:
Making progress. It's a small bug, but large complex fix

Issue #736, opened 2022-01-07
Hide suggestions and other tools from keyboard header


ZavenArra:
The text keyboard is supplying suggestions and other tools currently.  We do not require these tools and they should be hidden from the keyboard on all screens in this application

![Screenshot_20220107_145205_org greenstand android TreeTracker test](https://user-images.githubusercontent.com/364622/148617451-8025998c-fd59-457e-948e-589d95a36dad.jpg)

Elforama:
@ZavenArra I've looked at solutions for this online. There is no solution that seemed to fix this on all phones since all phones use different keyboard. Of the partial solutions I found, they all require the old XML to work. Compose doesn't seem to support the partial solutions.

ZavenArra:
Ah I see.  It's part of the specific implementation of Android on a given device or manufacturer.  Interesting.

ZavenArra:
Do you think we could file this as a feature request against compose and see if they are willing to work it out?

peakvalleytech:
Having you tried setting the input method type on the textfield?

On another, I don't think it is good UX to have the keyboard block the textfields. User should be able to see what they are typing.  I suggest we find a way to have the UI resize itself when the keyboard is shown.

Elforama:
We've tried that, but if you'd like to take a crack at fixing this issue feel free to let us know and I'll assign it to you.

PR #753, opened 2022-01-16
Hide suggestions

EngrTaofeek:
As requested in issue #736 I tried setting autofocus to false and it does not affect. see attached for documentation,
Screenshot 2022-01-16 at 16 22 46

The phone's keyboard settings overrides what ever option we put in our code.When I disabled autocomplete it went away also.
The other alternative which I did was to change the keyboard type to Password and not apply visual transformation so the text is visible.

this looks like this(notice it doesn't use the email keyboard) 
WhatsApp Image 2022-01-16 at 16 34 27
while it looked like this previously.
WhatsApp Image 2022-01-16 at 16 40 12


nationalkim (approved):
I'm seeing that this can look different depending on what device you use--for example, there is no microphone option on my phone, but it does add an extra row for numbers in its place. Just letting you know!

app/src/main/java/org/greenstand/android/TreeTracker/signup/CredentialEntryView.kt
```diff
@@ -115,7 +115,7 @@ fun CredentialEntryView(viewModel: SignupViewModel, state: SignUpState) {
                     onValueChange = { updatedEmail -> viewModel.updateEmail(updatedEmail) },
                     placeholder = { Text(text = stringResource(id = R.string.email_placeholder), color = Color.White) },
                     keyboardOptions = KeyboardOptions(
-                        keyboardType = KeyboardType.Email,
+                        keyboardType = KeyboardType.Password,
```
Elforama, line 118: I think we should leave this as email since it will make it much easier for people to type an email who have not used a phone before. Looking at solutions online, there doesn't seem to be a fix that works for all phones. Of the fixes I saw (while leaving it as email), they only work in XML, not compose.

Like Natalie said, different keyboards will show differently. I think the actual solution to this might be the user disabling suggestions on their keyboard.

I'd suggest turning off autocorrect for the keyboard.

Elforama (changes requested):
Please address comment.

app/src/main/java/org/greenstand/android/TreeTracker/signup/CredentialEntryView.kt
EngrTaofeek: Yes I thought as much, so we can close this PR without merging it.

EngrTaofeek:
> Please address comment.

I have addressed the requests in a new commit

Elforama (changes requested):
Please change password back to email.

PR #754, opened 2022-01-17
Update credential screen

### LINKED ISSUE #733: Email/Phone selection button is not to spec
The two button solution currently in the app does not follow the design closely enough.
- Buttons should be combined as in a switch
- Text should be bold
- Do not use gray color, use colors in spec
- Drop shadow should also makes use of vertical sides of button, as in the spec
- The widget in the app should look just like the spec

image

Gwen has offered to give additional support here.

### LINKED ISSUE #735: Move phone/email controls and text field to top of screen
The design shows the controls and text field on the phone/email page higher on the screen.  The app should look the same way.   The same holds for the name/organization page.   The keyboard should push the text field up so that the whole rectangle is above the keyboard and not just the text field line itself.   Apply this logic to ALL inputs.

Spec:
image

App Screenshot"
Screenshot_20220107_144529_org greenstand android TreeTracker test

EngrTaofeek:
fixes #735 fixes #733 Updated credential screen to fix issue #735 and designed credential buttons to spec, awaiting custom themes to be merged so I can apply it here likewise.
WhatsApp Image 2022-01-17 at 09 18 47


Elforama:
Before we merge this we need to confirm the design. The design we got didn't make sense.

EngrTaofeek:
> Before we merge this we need to confirm the design. The design we got didn't make sense.

Yes I agree, it was designed with just images and doesn't really consider the 3d effect and animation. We can raise it up in the mobile ux channel or Gwen

ZavenArra:
Gwen has confirmed the designs, this can be merged.

EngrTaofeek:
@Elforama based on the discussion yesterday, I was thinking of using the default depth button colours as used generally app and also add a bit of padding between them.

EngrTaofeek:
@Elforama Based on the discussions in the last meeting. I changed the default button to use the default depth button styles and added some padding between them. See screenshot below
WhatsApp Image 2022-01-24 at 12 59 54

PR #760, opened 2022-01-22
Approval buttons

### LINKED ISSUE #683: Update X and checkmark buttons
The selfie camera accept button is not square. A reusable square button should be used for both the selfie screen, tree capture screen, and note dialog.

There are two button types. A red X button, and a green check button. Both should be made reusable.

Selfie Screen: https://www.sketch.com/s/a0bbeac0-767a-4873-beef-c6c7ffd5826a/a/25ee5KQ/play
Tree Capture Screen: https://www.sketch.com/s/a0bbeac0-767a-4873-beef-c6c7ffd5826a/a/Gmddmgw/play

Files:
ImageReviewScreen.kt
TreeImageReviewScreen.tk

EngrTaofeek:
fixes #683 updated approval button to new icons and fixed bugs.
WhatsApp Image 2022-01-22 at 18 18 48-2
WhatsApp Image 2022-01-22 at 18 18 48


app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -86,6 +86,42 @@ fun BoxScope.ArrowButton(
     }
 }
 
+@Composable
+fun AcceptButton(
```
Elforama, line 90: Let's try to have less repeated code here. The contents of Accept and Decline are the same aside from color and icon. Can we make a generic button that accepts those two params? Then these two buttons can just reuse that button with different icon/colors.

This will allow the buttons to stay in sync.

Currently we need to keep two buttons in sync. Someone can update one and forget about the other.
If they both reuse a single button, they will always stay in sync.

app/src/main/res/layout/activity_main.xml
```diff
@@ -26,6 +26,7 @@
                 android:layout_width="wrap_content"
                 android:layout_height="wrap_content"
                 android:layout_gravity="center"
+                android:textColorLink="#430404"
```
Elforama, line 29: What is this for?

app/src/main/java/org/greenstand/android/TreeTracker/camera/ImageReviewScreen.kt
```diff
@@ -22,6 +22,7 @@ import androidx.compose.ui.layout.ContentScale
 import androidx.compose.ui.platform.LocalContext
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.unit.dp
+import kotlinx.coroutines.launch
```
Elforama, line 25: Do we need this?

Elforama (changes requested):
Looks good, just some minor cleanup requested.

app/src/main/java/org/greenstand/android/TreeTracker/camera/ImageReviewScreen.kt
EngrTaofeek: Yes it's been used in the accept button to approve images in the view model.

app/src/main/res/layout/activity_main.xml
EngrTaofeek: Not needed. that was a mistake. I have reverted it in a more recent commit.

PR #762, opened 2022-01-25
Update user buttons

### LINKED ISSUE #726: Fix user button image corners
The corner of the user image on the user image button (e.g. https://www.sketch.com/s/a0bbeac0-767a-4873-beef-c6c7ffd5826a/a/4a55aKo) does not match the rounded edge of the button shape behind it. Please make the image have the same rounded edge.

### LINKED ISSUE #740: Grower selection buttons not constructed to spec
image

image

EngrTaofeek:
fixes #740    Fixed border corner radius and allowed depth button outline show above the image. And also background. fixes #726 
WhatsApp Image 2022-01-26 at 00 39 27


EngrTaofeek:
@Elforama I had a challenge implementing the shadow effect for a selected grower button as shown here https://www.sketch.com/s/a0bbeac0-767a-4873-beef-c6c7ffd5826a/a/4a55aKo  The best way to create this effect is to use radial gradients. However radial gradients don't quite work well with other shapes except if they are circular or oval, see what it looks like(it takes the oval shape).
WhatsApp Image 2022-01-26 at 10 23 04
Other available options were to use another external library(https://github.com/PatilSiddhesh/Holi) or spend more time trying to create the effect from the scratch.
I explored with other gradients that works perfectly. Just in case you want me to use another one instead of spending more time on that
WhatsApp Image 2022-01-26 at 10 23 48
WhatsApp Image 2022-01-26 at 10 24 22
WhatsApp Image 2022-01-26 at 10 25 08
WhatsApp Image 2022-01-26 at 10 25 46
WhatsApp Image 2022-01-26 at 10 26 58
WhatsApp Image 2022-01-26 at 10 28 08

@Elforama and @snacks what do you think?

app/src/main/java/org/greenstand/android/TreeTracker/view/SelectableItemDetail.kt
```diff
@@ -62,7 +53,7 @@ fun SelectableImageDetail(
                     .fillMaxWidth()
                     .wrapContentHeight()
                     .aspectRatio(1.0f)
-                    .clip(RoundedCornerShape(percent = 10)),
+                    .clip(RoundedCornerShape( 10.dp)),
```
Elforama, line 56: <nit> please don't add extra spacing for params

Elforama (changes requested):
I think the modifications to this prior to this PR where to add a green outline when selected. I think we can make this a default button without any glow and 'if' statements. It can work much like the language selection screen.

Once that is cleaned up we can then make a 'glow' modifier. Something that way we just add a glow and the whatever view we want can start to glow.

EngrTaofeek:
> I think the modifications to this prior to this PR where to add a green outline when selected. I think we can make this a default button without any glow and 'if' statements. It can work much like the language selection screen.
> 
> Once that is cleaned up we can then make a 'glow' modifier. Something that way we just add a glow and the whatever view we want can start to glow.

Okay great. I haven't added that green outline feature to this branch yet. I have fixed the extra spacing you pointed out and rebased the branch.

Elforama:
Can you please attach a new screenshot that shows the current state?

EngrTaofeek:
@Elforama  it looks like this now (default and selected state)
WhatsApp Image 2022-02-02 at 00 19 54

WhatsApp Image 2022-02-02 at 00 19 21

PR #764, opened 2022-01-28
CodeCleanup in splashScreen

sidsharma2002:


app/src/main/java/org/greenstand/android/TreeTracker/splash/SplashScreen.kt
```diff
@@ -80,6 +71,25 @@ fun SplashScreen(
     )
 }
 
+fun isFineAndCoarseLocationPermissionGiven(result: Map<String, Boolean>): Boolean {
```
Elforama, line 74: Please make these 3 methods private.

app/src/main/java/org/greenstand/android/TreeTracker/splash/SplashScreen.kt
```diff
@@ -36,26 +36,17 @@ fun SplashScreen(
         contract = ActivityResultContracts.RequestMultiplePermissions(),
         onResult = { result ->
             scope.launch {
-                if (result[Manifest.permission.ACCESS_FINE_LOCATION] == true &&
-                    result[Manifest.permission.ACCESS_COARSE_LOCATION] == true
-                ) {
+                if (isFineAndCoarseLocationPermissionGiven(result)) {
```
Elforama, line 39: I think a less verbose name would do, such as 'isLocationPermissionGranted'

app/src/main/java/org/greenstand/android/TreeTracker/splash/SplashScreenViewModel.kt
```diff
@@ -9,9 +9,9 @@ class SplashScreenViewModel(
     private val users: Users,
 ) : ViewModel() {
 
-    fun migratePreferences() {
+    fun migratePreferencesIfNeeded() {
         preferencesMigrator.migrateIfNeeded()
     }
 
-    suspend fun requiresInitialSetup(): Boolean = users.getPowerUser() == null
+    suspend fun doesRequireInitialSetup(): Boolean = users.getPowerUser() == null
```
Elforama, line 16: I'm guessing you updated this to sound more like it returns a boolean, but I think it still sounds like a statement. Can reword this to start with "is"?

Elforama (changes requested):
Thanks for taking the initiative to clean things up! This is looking great, I just added a few comments.

app/src/main/java/org/greenstand/android/TreeTracker/splash/SplashScreenViewModel.kt
sidsharma2002: what about 'isInitialSetupRequired' ?

Elforama (approved):
Thanks for the changes!

PR #767, opened 2022-01-31
Generic dialog

EngrTaofeek:
Generic dialog that can be used for all scenarios across the app and build different types of dialogs depending on parameters filled. Check out screenshots 
WhatsApp Image 2022-01-31 at 21 29 52-2
WhatsApp Image 2022-01-31 at 21 29 52-3
WhatsApp Image 2022-01-31 at 21 29 52
below for sample usages


app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -113,6 +118,112 @@ fun ApprovalButton(
...
+    dialogIcon: Painter = painterResource(id = R.drawable.greenstand_logo),
+    title: String = "",
+    content: String = "",
+    onPositiveClick: () -> Unit = {},
+    onNegativeClick: () -> Unit = {},
+    isTextInputAvailable: Boolean = false,
+    textInputValue: String = "",
+    onTextInputValueChange: (String) -> Unit = {},
+    showPositiveButtonDialog: Boolean = true,
+    showNegativeButtonDialog: Boolean = false,
```
Elforama, line 143: We can use null onPositive/onNegativeClick params to determine if we show the buttons. 
e.g. If we want to positive only button, we only pass that click listener. Since the negative click listener is null, it is turned off.

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -113,6 +118,112 @@ fun ApprovalButton(
...
+         */
+fun CustomDialog(
+    dialogIcon: Painter = painterResource(id = R.drawable.greenstand_logo),
+    title: String = "",
+    content: String = "",
+    onPositiveClick: () -> Unit = {},
+    onNegativeClick: () -> Unit = {},
+    isTextInputAvailable: Boolean = false,
+    textInputValue: String = "",
+    onTextInputValueChange: (String) -> Unit = {},
```
Elforama, line 141: Same thing as below. Instead of passing a boolean, we can turn it on/off depending if a click listener is added.

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -113,6 +118,112 @@ fun ApprovalButton(
...
+                    style = CustomTheme.typography.medium,
+                    fontWeight = FontWeight.Bold,
+                )
+            }
+        },
+        modifier = Modifier
+            .fillMaxWidth()
+            .wrapContentHeight()
+            .padding(2.dp)
+            .border(1.dp, color = Color(0xFF75B926), shape = RoundedCornerShape(percent = 10))
```
Elforama, line 174: Please use app colors.

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -113,6 +118,112 @@ fun ApprovalButton(
...
+            .clip(RoundedCornerShape(percent = 10)),
+        backgroundColor = Color(0XFF191C1F),
```
Elforama, line 176: Please use app colors.

Elforama (changes requested):
Looking good. I've left a few comments to hopefully reduce the number or params.

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -171,26 +163,26 @@ fun CustomDialog(
...
             ) {
-
                 Text(
                     text = content,
                     color = CustomTheme.textColors.primaryText,
                     style = CustomTheme.typography.regular,
                     modifier = Modifier.padding(bottom = 5.dp)
                 )
-                if (isTextInputAvailable) {
+                onTextInputValueChange?.let {
```
Elforama, line 181: <nit> when a 'let' block is large like this 'it' can lose context. I'd suggest to give 'it' a name since it's used 4 lines below.

Elforama (approved):
Thanks for the fixes. Looks good.

PR #769, opened 2022-02-06
Updating dialogs to custom

### LINKED ISSUE #728: Create a generic treetracker styled dialog
All our dialogs are placeholders until design makes something. For the time being we need to make a generic dialog that uses the depth button which can be reused for all our use cases.

Something like:
TreeTrackerDialog(onPositiveAction: () -> Unit, onNegativeAction: () -> Unit, content: @Composable () -> Unit)

EngrTaofeek:
fixes #728  Updated all previous dialogs to use new custom dialogs. And made minor refactoring on the custom dialog and TreeImageReviewViewModel to correctly save note and update it. They are now functional.
WhatsApp Image 2022-02-06 at 11 23 07-2
WhatsApp Image 2022-02-06 at 11 23 07


Elforama:
This doesn't seem to work for the user already exist dialog. Can we make the dialog more customizable like the android dialogs are? Then these other dialogs can be built upon it.

Elforama (changes requested):
Please support all dialogs with new dialog system.

EngrTaofeek:
> This doesn't seem to work for the user already exist dialog. Can we make the dialog more customizable like the android dialogs are? Then these other dialogs can be built upon it.

You are right I forgot about that.

Elforama (changes requested):
Thanks for the fix. I left some minor comments. Once addressed I'll merge.

Elforama (approved):

PR #772, opened 2022-02-13
Updating dialogs to custom with fixes after revert (#769)

Elforama:
* Updated bad location dialog to use new custom dialogs

* Applied custom dialog to permissions screen and TreeImageReviewScreen.kt. Little refactoring on viewmodel.

* Updated CustomDialog to take in composables and display them. Also refactored content to textContent

* Applied Custom dialog to user already exist dialog

* Requested PR changes


EngrTaofeek (approved):
Looks good

PR #788, opened 2022-02-19
Fix sessionId crash

Elforama:


app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureViewModel.kt
```diff
@@ -43,6 +45,7 @@ class TreeCaptureViewModel(
     }
 
     suspend fun endSession() {
+        locationDataCapturer.stopGpsUpdates()
```
EngrTaofeek, line 48: I am not sure if sessionID is used in Tree Image Review would be better to call this in there in that case

EngrTaofeek (changes requested):
Looks great starting the session tracking from the wallet screen and ending after treecapturing. Refactoring the locations.kt file is great and made it much readable. Unresolved reference: Convergence somewhere in the project is causing it to fail the test, this was probably introduced by the refactor. That's the only thing to be fixed in this PR.

ZavenArra:
I can also confirm that this PR fixes the sessionId crash

app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureViewModel.kt
Elforama: I'm not quite sure what you mean.

This is stopping GPS updates when the user ends the session. It can only be ended from the tree capture screen.

PR #789, opened 2022-02-19
Build variant fix

### LINKED ISSUE #783: SessionTracker bug
In the release version, immediately after capturing trees and returning to the Dashboard, the app crashes due to sessionID being null. NOTE: this doesn't occur in the debug version.
Here is a link to the issue on firebase Crashlytics https://console.firebase.google.com/u/0/project/tree-tracker-a7a13/crashlytics/app/android:org.greenstand.android.TreeTracker.test/issues/f6be56f6c80dfc9e4f044b2dccc9ffef?time=last-seven-days&sessionEventKey=620DF83C00BE0001715BB359B8D92043_1644022452200519059

### LINKED ISSUE #784: Crashing while uploading in release version
In release version, clicking on the upload button crashes the app. It sends the following to the Crashlytics console https://console.firebase.google.com/u/0/project/tree-tracker-a7a13/crashlytics/app/android:org.greenstand.android.TreeTracker.test/issues/6ebf2cf1f0d2d3718cd5f816ec19d8f7?time=last-seven-days&sessionEventKey=620DF929009700017DB5B359B8D92043_1644022958965512895

### LINKED ISSUE #786: Significance difference between Debug and Firebase Release output
I generated a release signed APK from android studio and noticed there was a significant difference with the release version available on firebase other than what was explicitly programmed to be different especially in terms of performance. Solving this should also solve #783 and issue #784

EngrTaofeek:
fixes #783. fixes #784. fixes #786  Removed applicationIdSuffix from build variants other than release and debug to enable it perform well.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -95,7 +95,6 @@ android {
 
         prerelease {
             initWith(debug)
-            applicationIdSuffix ".prerelease"
             buildConfigField "String", "OBJECT_STORAGE_IDENTITY_POOL_ID", "\"${s3_production_identity_pool_id}\""
             buildConfigField 'String', 'OBJECT_STORAGE_ENDPOINT', '"eu-central-1"'
             buildConfigField 'String', 'OBJECT_STORAGE_BUCKET_IMAGES', '"treetracker-production-images"'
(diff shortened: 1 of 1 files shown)


ZavenArra (changes requested):
Unfortunately, I don't think we can accept this change.

We need to use the application suffix so we can install both the beta and production versions.

PR #797, opened 2022-02-21
Added glow for 2.0

EngrTaofeek:
I added the glow feature for 2.0 as requested by Zaven during our conversation. He said that should be fine for a temporary fix.

 fixes #762

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/view/SelectableItemDetail.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/view/SelectableItemDetail.kt
@@ -1,6 +1,7 @@
 package org.greenstand.android.TreeTracker.view
 
 import androidx.compose.foundation.Image
+import androidx.compose.foundation.border
 import androidx.compose.foundation.layout.Column
 import androidx.compose.foundation.layout.Row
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/view/SelectableItemDetail.kt
```diff
@@ -39,14 +41,25 @@ fun SelectableImageDetail(
...
-            .padding(8.dp)
-            .width(156.dp)
-            .clip(RoundedCornerShape(10.dp))
-            .wrapContentHeight(),
+        modifier = if (isSelected) userButtonModifier
+            .border(
+                width = 1.dp, brush = Brush.verticalGradient(
+                    colors = listOf(
+                        AppColors.Gray,
+                        AppColors.Green
```
Elforama, line 58: I think it might make more sense to use solid green, no gradient, or use a darker green instead of gray.

Elforama (approved):
Looks good. Just not sure if there should be more green.

ZavenArra:
Tested this PR and accepting it as good enough for now.  We can work toward a real 'glow' in 2.2 or so

PR #798, opened 2022-02-22
Made recapturing easier without leaving TreeCaptureScreen

### LINKED ISSUE #795: Retry location capture when no convergence
If the GPS cannot converge, after the user is given the GPS error dialog, a positive response (thumbs up button) should attempt to reconverge the location rather than popping the stack.

Start here: https://github.com/Greenstand/treetracker-android/pull/794

EngrTaofeek:
Made recapturing easier without leaving TreeCaptureScreen. Previously clicking the positive approval button takes you back to the walletSelectScreen instead. fixes #795


Elforama:
I'm not quite sure what this fixes.

So if you have no GPS, clicking the green button kicks the user back to the wallet select screen, but now it leaves them on the screen to try capturing again?

Elforama (approved):

Issue #799, opened 2022-02-22
alternative return paths to dashboard do not update capture count


ZavenArra:
To reproduce:

1. Enter tracking mode, select a user, select a target wallet
2. Track one tree
3. Track another tree
4. Click the user icon to change users
5. Click the finger pointing back button
**6. Trees not added to upload countdown bar (total is 0 but should be 2)**
7.  Enter tracking mode again, select a user, select a target wallet
8. Track one tree
9. Track another tree
10. Exit tracking mode by click the finger pointing back button 
**11. Upload countdown bar now shows 4 trees (from both sessions)**

Elforama:
What path?

Elforama:
Are you testing this with the fake tree planting button?

Elforama:
I think this feature is working correctly. It's just that legacy trees are considered uploaded by a different status which is messing with the UI. If you only plant manually it should be working (only 2.0 trees). Once users upload all legacy trees the UI will have no more issues.

ZavenArra:
Verified this issue is not related to the legacy trees.  Update with steps to reproduce.

PR #814, opened 2022-02-23
Update progress bar for edge cases

Elforama:
fixes https://github.com/Greenstand/treetracker-android/issues/799

UserSelect screen went back to dashboard by popping backstack which avoided a dashbaord refresh. It was fixed by navigating to it instead of popping backstack. This is not ideal, but it works for now, and I don't see other issues due to it.


ZavenArra:
verified fixed.

PR #817, opened 2022-02-24
Add temp org screen

Elforama:
This adds a temporary screen between selecting a wallet and planting trees to allow adding an org for the session.

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/models/NavRoute.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/models/NavRoute.kt
@@ -14,6 +14,7 @@ import org.greenstand.android.TreeTracker.capture.TreeImageReviewScreen
 import org.greenstand.android.TreeTracker.dashboard.DashboardScreen
 import org.greenstand.android.TreeTracker.languagepicker.LanguageSelectScreen
 import org.greenstand.android.TreeTracker.messages.MessagesUserSelectScreen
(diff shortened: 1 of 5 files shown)


ZavenArra:
Verified working, and verified the org field is populated in the bulk pack.  However need once change, the organization needs to be 'sticky' - we can expect our users to type it in each time or remember it.

ZavenArra (changes requested):
The feature is working well.  We need one addition - the organization needs to be sticky and pre-populate the input field with the most recent value that was entered.

In general our users often will not remember the specific organization name they are given, so it's important that this is saved the first time time it is entered.  We may need to label the field in this case well, since if it is prepopulated it might not be clear what the field is for.

EngrTaofeek:
@ZavenArra Another option which is more of a permanent solution to have a list of organisations displayed like those language buttons and make them selectable. It solves both it been sticky and prevents wrong spellings.

EngrTaofeek (approved):
Other than Zaven's request, this is good.

Elforama:
@ZavenArra What if I add the button used in the language screen (style wise) to act as a "sticky" autofill. Pressing it autofills the selection, or can move onto the next screen. This would leave the form untouched with the default hint. When there was no previous org, the button would not show.

@EngrTaofeek Our long term solution is to not allow users to add orgs manually, so inputting will not work. We'll be using links to auto add orgs to an org screen that's accessible from the dashboard.

PR #819, opened 2022-02-24
Updated KOIN dependencies

EngrTaofeek:
Updated KOIN dependencies to the recent ones more suitable for compose and fixed imports related errors. Also updated navigation-compose to be compatible with that version of KOIN.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -207,11 +207,7 @@ dependencies {
     implementation 'com.amazonaws:aws-android-sdk-s3:2.16.8'
 
     //koin dependencies
-    implementation "io.insert-koin:koin-core:$koin_version"
     implementation "io.insert-koin:koin-android:$koin_version"
-    implementation "io.insert-koin:koin-android-compat:$koin_version"
-    implementation "io.insert-koin:koin-androidx-workmanager:$koin_version"
-    implementation "io.insert-koin:koin-androidx-navigation:$koin_version"
     implementation "io.insert-koin:koin-androidx-compose:$koin_version"
     testImplementation "io.insert-koin:koin-test:$koin_version"
 
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -206,7 +206,14 @@ dependencies {
     implementation 'com.amazonaws:aws-android-sdk-core:2.16.8'
     implementation 'com.amazonaws:aws-android-sdk-s3:2.16.8'
 
-    implementation 'org.koin:koin-android-viewmodel:2.0.0-GA'
+    //koin dependencies
+    implementation "io.insert-koin:koin-core:$koin_version"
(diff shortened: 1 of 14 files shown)


Elforama (changes requested):
Please undo start imports

app/build.gradle
```diff
@@ -206,7 +206,14 @@ dependencies {
     implementation 'com.amazonaws:aws-android-sdk-core:2.16.8'
     implementation 'com.amazonaws:aws-android-sdk-s3:2.16.8'
 
-    implementation 'org.koin:koin-android-viewmodel:2.0.0-GA'
+    //koin dependencies
+    implementation "io.insert-koin:koin-core:$koin_version"
+    implementation "io.insert-koin:koin-android:$koin_version"
+    implementation "io.insert-koin:koin-android-compat:$koin_version"
```
Elforama, line 212: compat is for Java, so we shouldn't need it

app/build.gradle
```diff
@@ -206,7 +206,14 @@ dependencies {
...
+    implementation "io.insert-koin:koin-androidx-workmanager:$koin_version"
```
Elforama, line 213: Unless there are issues with our current usage of Koin in our workmanager, I don't think we need this.

app/build.gradle
```diff
@@ -206,7 +206,14 @@ dependencies {
...
+    implementation "io.insert-koin:koin-androidx-navigation:$koin_version"
```
Elforama, line 214: do you know what this navigation module adds? I can't find any documentation on it.

app/build.gradle
```diff
@@ -206,7 +206,14 @@ dependencies {
     implementation 'com.amazonaws:aws-android-sdk-core:2.16.8'
     implementation 'com.amazonaws:aws-android-sdk-s3:2.16.8'
 
-    implementation 'org.koin:koin-android-viewmodel:2.0.0-GA'
+    //koin dependencies
+    implementation "io.insert-koin:koin-core:$koin_version"
+    implementation "io.insert-koin:koin-android:$koin_version"
```
Elforama, line 211: Do we need both of these? I'm curious if the android version contains core in it already

app/build.gradle
EngrTaofeek: Okay I didn't know that.

app/build.gradle
EngrTaofeek: Not both. I just realised Koin android alone is enough for the project.

Elforama:
I think you might have to rebase. You have other people commits showing up.

EngrTaofeek:
> I think you might have to rebase. You have other people commits showing up.

I rebased that's why it added the commits merged after I created the branch.

PR #830, opened 2022-02-26
Tutorial and guides

### LINKED ISSUE #738: Need a tutorial image for selfie and image capture screens
image

EngrTaofeek:
fixes #738 Implemented Selfie, Capture and ImageReview tutorial dialogs in the app to allow new app installs see tutorial and people who click info button see it also. Also made a bit of refactor in the Selfie screen.


ZavenArra:
Testing this one out now..

ZavenArra (approved):
Ran through this PR on the phone, looks great.

app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeImageReviewViewModel.kt
```diff
@@ -27,12 +35,19 @@ class TreeImageReviewViewModel(
...
+    }
+
     fun addNote(){
         viewModelScope.launch {
             treeCapturer.setNote(_state.value!!.note)
             _state.value = _state.value?.copy(isDialogOpen = false)
         }
     }
+
+    suspend fun isFirstTrack(): Boolean = users.getPowerUser()!!.numberOfTrees < 1
```
Elforama, line 49: This should be private

app/src/main/java/org/greenstand/android/TreeTracker/signup/SignupViewModel.kt
```diff
@@ -116,6 +117,12 @@ class SignupViewModel(
         )
     }
 
+    fun updateSelfieTutorialDialog(state: Boolean){
+        _state.value = _state.value?.copy(showSelfieTutorial = state)
+    }
+
+    suspend fun isInitialSetupRequired(): Boolean = users.getPowerUser() == null
```
Elforama, line 124: Should be private

app/src/main/java/org/greenstand/android/TreeTracker/view/Tutorial.kt
```diff
@@ -0,0 +1,249 @@
...
+                    contentDescription = null
+                )
+            }
+        }
+
+    }
+}
+
+@Composable
+fun TutorialDialog(
```
Elforama, line 217: This is fine for now. But ideally you'd reuse the dialog you created before. 

If you wanted to update the dialog style with your current implementation, you'd have to remember to update it in two locations instead of one.

We should have our own version of an "AlertDialog" that we can reuse.

Elforama (commented):
Awesome design work! I've left some feedback for you to think about.

PR #831, opened 2022-02-26
Updated FeatureFlags in 

EngrTaofeek:
Updated FeatureFlags.kt to fix dashboard in beta release.
it currently looks like this(bug introduced by #809 ).

this PR returns it bACK TO THIS


Elforama:
@EngrTaofeek Perhaps I'm missing something, but I don't see how this changes any of the logic. It's still checking if it's BETA

app/src/main/java/org/greenstand/android/TreeTracker/models/FeatureFlags.kt
```diff
@@ -5,9 +5,9 @@ import org.greenstand.android.TreeTracker.BuildConfig
 object FeatureFlags {
     val DEBUG_ENABLED: Boolean = (
         BuildConfig.BUILD_TYPE == "dev" ||
-            BuildConfig.BUILD_TYPE == "debug" ||
-            BuildConfig.BUILD_TYPE == "beta"
```
EngrTaofeek, line 9: @Elforama beta variant here was added as debug enabled. so in the Dashboard since it is debug enabled it has both the Upload button and progress bar on the same row. Then since the beta doesn't have a message feature flag it becomes empty and messes up the screen. Zaven made some changes and merged them himself that's probably you missed it. Check the DashboardScreen.kt file or the merged PR #809

PR #864, opened 2022-03-02
Apply old image rotation logic

Elforama:
This SHOULD fix the image rotation. But it cannot be confirmed until we test it on devices with the image rotation bug. We can merge and push it to beta. Once it's confirmed by testers with image issues to be fixed, we can close the task.

This solution is not optimal since we are rotating the image twice... once in resize function, then again in orient function. Ideally we'd modify the image completely in one pass, but that can be optimized later.


ZavenArra:
It appears that the correct solution is simply to add the exif rotation data back into the image file, and remove all the other various rotation logic that was incorrectly accounting for this.

I believe that I now remember that 90 degrees is most often the default rotation associated with a portrait image, but some devices use 0.  This leads to a lot fo confusion.

PR #867, opened 2022-03-02
Made Custom snackbar and implemented validation error messages in them.

### LINKED ISSUE #859: Snack bar for phone or email validation error

EngrTaofeek:
Fixes #859 Made Custom snackbar and implemented validation error messages in them.
I think the previous refactor of composable into EmailTextField is a bit unnecessary as it uses a lot of values in CredentialEntry composable now the function takes in 6 parameters that are constants.

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -24,7 +24,7 @@ android {
         applicationId "org.greenstand.android.TreeTracker"
         minSdkVersion 21
         targetSdkVersion 31
-        versionCode 145
+        versionCode 146
         versionName "1.4.0"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         multiDexEnabled true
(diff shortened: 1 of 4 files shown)


ZavenArra:
When I click that done button, the keyboard does not hide, so I do not see the custom message.   Can we put the message above the keyboard, or hide the keyboard when the done typing button is tapped.

ZavenArra (changes requested):
When I click that done button, the keyboard does not hide, so I do not see the custom message. Can we put the message above the keyboard, or hide the keyboard when the done typing button is tapped.

Elforama:
What if we show the error on the text input where the input goes red?

EngrTaofeek:
> What if we show the error on the text input where the input goes red?

We would also have to show the error message outside the inbox field so the user knows what to do.

ZavenArra (approved):

PR #889, opened 2022-03-07
Check for GPS device in the phone and show dialog if not present.

### LINKED ISSUE #876: Detect absence of GPS unit
Some phones doesn't have a GPS unit

The older version of the app has some logic to detect that there is no GPS unit.   This can be determined by asking the location service which providers are available.

https://github.com/Greenstand/treetracker-android/tree/release

EngrTaofeek:
fix #876 Check for GPS device in the phone and show dialog if not present. Clicking the button in the dialog closes the app.
Screenshot_20220307-191728

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/activities/TreeTrackerActivity.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/activities/TreeTrackerActivity.kt
@@ -1,15 +1,26 @@
 package org.greenstand.android.TreeTracker.activities
 
+import android.app.PendingIntent.getActivity
+import android.content.Context
+import android.location.LocationManager
 import android.os.Bundle
 import androidx.activity.ComponentActivity
 import androidx.activity.compose.setContent
+import androidx.compose.material.MaterialTheme
+import androidx.compose.material.Surface
+import androidx.compose.runtime.Composable
 import androidx.compose.runtime.ExperimentalComposeApi
+import androidx.compose.ui.res.painterResource
+import androidx.compose.ui.res.stringResource
+import org.greenstand.android.TreeTracker.R
 import org.greenstand.android.TreeTracker.models.FeatureFlags
 import org.greenstand.android.TreeTracker.models.Language
 import org.greenstand.android.TreeTracker.models.LanguageSwitcher
 import org.greenstand.android.TreeTracker.models.TreeTrackerViewModelFactory
 import org.greenstand.android.TreeTracker.root.Root
 import org.greenstand.android.TreeTracker.theme.CustomTheme
+import org.greenstand.android.TreeTracker.view.AppColors
+import org.greenstand.android.TreeTracker.view.CustomDialog
 import org.koin.android.ext.android.inject
 
 class TreeTrackerActivity : ComponentActivity() {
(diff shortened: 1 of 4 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/activities/TreeTrackerActivity.kt
```diff
@@ -26,11 +37,36 @@ class TreeTrackerActivity : ComponentActivity() {
...
+                }
             }
         }
     }
+
+    fun hasGPSDevice(context: Context): Boolean {
+        val mgr = context.getSystemService(LOCATION_SERVICE) as LocationManager
+        val providers = mgr.allProviders
+        return providers.contains(LocationManager.GPS_PROVIDER)
+    }
```
Elforama, line 61: Let's try to keep the activity as clean as possible. It's easy for activities to turn into super large classes as time goes on. We can move this logic into one of the Location classes we have.

app/src/main/java/org/greenstand/android/TreeTracker/activities/TreeTrackerActivity.kt
```diff
@@ -26,11 +37,36 @@ class TreeTrackerActivity : ComponentActivity() {
...
+}
+
+@Composable
+fun NoGPSDeviceDialog(onPositiveClick: () -> Unit){
+    CustomDialog(
+        dialogIcon = painterResource(id = R.drawable.error_outline),
+        title = stringResource(R.string.no_gps_device_header),
+        textContent = stringResource(R.string.no_gps_device_content),
+        onPositiveClick = onPositiveClick
+    )
```
Elforama, line 71: Please move this to one of the dialog files, or create a new one.

app/src/main/java/org/greenstand/android/TreeTracker/activities/TreeTrackerActivity.kt
```diff
@@ -26,11 +37,36 @@ class TreeTrackerActivity : ComponentActivity() {
         } else {
             languageSwitcher.applyCurrentLanguage(this)
         }
-
-        setContent {
-            CustomTheme {
-                Root(viewModelFactory)
+        if (hasGPSDevice(this)) {
```
Elforama, line 40: It looks like the main difference here is Root vs NoGPSDiviceDialog. We can move the if statement inside the custom theme.

app/src/main/res/values/strings.xml
```diff
@@ -104,4 +104,6 @@
     <string name="selfie_instruction">to take a selfie to continue your registration.</string>
     <string name="recapture_tutorial">Click on the button to try capturing the tree again.</string>
     <string name="capture_tutorial">Click on the button to save tree.</string>
+    <string name="no_gps_device_header">NO GPS DEVICE FOUND</string>
+    <string name="no_gps_device_content">This app is not supported on this device because no GPS receiver was found. Kindly try another phone that has GPS device</string>
```
Elforama, line 108: I think we can reword this to say something more like 
"This device does not support GPS. Please use a device that has GPS to continue."

app/src/main/res/values/strings.xml
```diff
@@ -104,4 +104,6 @@
     <string name="selfie_instruction">to take a selfie to continue your registration.</string>
     <string name="recapture_tutorial">Click on the button to try capturing the tree again.</string>
     <string name="capture_tutorial">Click on the button to save tree.</string>
+    <string name="no_gps_device_header">NO GPS DEVICE FOUND</string>
```
Elforama, line 107: This can probably say  "No GPS found on device"

Elforama (changes requested):
Looks good, but needs some cleanup, and possible rewording of the dialog.

app/src/main/java/org/greenstand/android/TreeTracker/models/location/Locations.kt
```diff
@@ -27,3 +51,9 @@ data class LocationData(
     val convergenceStatus: ConvergenceStatus?,
     val capturedAt: String
 )
+
+fun hasGPSDevice(context: Context): Boolean {
```
Elforama, line 55: Please move this into a util class to either use statically, or to inject

Elforama (changes requested):
Looks good. Please just address the comment about the GPS function

app/src/main/res/values/strings.xml
ZavenArra: We also need a workflow for making sure we have a swahili translation for any new text added.  @sebastiangaertner can provide this.  Please file a ticket for him.

app/src/main/res/values/strings.xml
ZavenArra: same here, Swahili needed @sebastiangaertner

app/src/main/res/values/strings.xml
```diff
@@ -104,4 +104,6 @@
     <string name="selfie_instruction">to take a selfie to continue your registration.</string>
     <string name="recapture_tutorial">Click on the button to try capturing the tree again.</string>
     <string name="capture_tutorial">Click on the button to save tree.</string>
+    <string name="no_gps_device_header">No GPS found on device</string>
+    <string name="no_gps_device_content">This device does not support GPS. Please use a device that has GPS to continue.</string>
```
ZavenArra, line 108: Always need a swahili translation of any new text added.

EngrTaofeek:
This PR has been moved to #967

PR #891, opened 2022-03-07
Made plant identifier to be case insensitive 

### LINKED ISSUE #871: Planter identifier should be case insensitive
lowercase all text entered in the 'email' before storing to the internal mobile database or checking for pre-existing wallet registrations.

EngrTaofeek:
fixes #871  Changed Keyboard for email entry to lowercase as default option. Also Converts email to lowercase and make visible for the user instantly(In case user decides to switch to Caps he/she would only see small letters in the Textfield). No Input Filters for Compose so this was the best I could come up with.


Elforama (changes requested):
This looks great. I noticed an issue with the AddWallet screen. It still lets my type capitol letters. I think you do use the same lowercase logic you did on the signup screen.

Elforama (approved):

PR #900, opened 2022-03-09
fix re importing drawables : white_leaf.png and yellow_leafs_placehold…

Tarek-Bohdima:
…er.png

with different densities


Elforama:
Is it possible to import this asset as an SVG to simplify the assets? Otherwise, this looks good

ZavenArra:
One user is still experiencing what appears to be this crash, on a Samsung J3 Galaxy.

PR #903, opened 2022-03-10
Individual message list

EngrTaofeek:
Created Individual message list with dummy data.
Would push a commit with a data class after the meeting to update the MessageItemButton well
Screenshot_1646930856


Elforama:
A few notes to help moving forward with this using the message api.
- You can use a 'when' statement in the UI to make different list items based on the data type
- When clicking on an item to open it, we can pass the id of that item to the next screen. The other screen can use that id to load the info for it
- We can use a viewmodel factory (look at AddWalletScreen) to pass in an id into the viewmodel without call a method within a 'launchEffect'

app/src/main/java/org/greenstand/android/TreeTracker/view/SelectableItemDetail.kt
```diff
@@ -30,8 +33,9 @@ import org.greenstand.android.TreeTracker.theme.CustomTheme
 
 @Composable
 fun SelectableImageDetail(
-    photoPath: String,
+    photoPath: String? = null,
     isSelected: Boolean,
+    painter: Painter? = null,
```
Elforama, line 38: Since now have multiple image types, we should avoid passing conflicting data and instead extract the image outside, and have the caller pass it in as another composable.

Issue #912, opened 2022-03-13
Upload images async


Elforama:
We currently upload one image at a time. We can possibly speed this up my uploading multiple at once.

PR #914, opened 2022-03-13
upload images async

Elforama:
fixes https://github.com/Greenstand/treetracker-android/issues/912


ZavenArra (approved):
Tested this out on a few phones, and seems good!

Issue #828, opened 2022-02-26
Should a language be initially selected when the user first starts the app?


ZavenArra:

Elforama:
Based on Gwen's comment in slack we should pre select the language the device is set to. If it's not English or Swahili I think we should default to English since it's more universal.

PR #918, opened 2022-03-14
no language default

ZavenArra:
- wip: attempt no language default
- feat: implement no default language

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/languagepicker/LanguagePickerViewModel.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/languagepicker/LanguagePickerViewModel.kt
@@ -17,7 +17,7 @@ class LanguagePickerViewModel(
 ) : ViewModel() {
 
     private val _currentLanguage = MutableLiveData(languageSwitcher.currentLanguage())
-    val currentLanguage: LiveData<Language> = _currentLanguage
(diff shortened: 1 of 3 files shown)


Elforama (changes requested):
Please see comment above

ZavenArra:
I believe @snacks's determination was that there would be no initial language default.  That's also what Karen was confused about - the initial language selection screen.

Elforama:
@ZavenArra Im confused -> https://github.com/Greenstand/treetracker-android/issues/828

ZavenArra:
I am also confused.

Also one of the main use cases we have heard about in the past is that users will have devices that are set to english, but actually want to see Swahili in the app but don't know how to change this universally in the settings right when they are using the device. 

The other issue is that some of our testers are not clear on what to do on the language selection screen.

snacks:
Yes I see the confusion here, which is mostly on me. Sorry about that! The initial assumption was that users would know a language button was already selected when they come to that screen but testing is showing that to not be true. Reading the above case that phones could be initially set to English but the user would prefer Swahili reinforces this decision to not have a "default" language.


I think a better way to describe this ticket would be that we want there to be "no selected language" on the language selection screen the first time the user visits. When they re-visit the language selection screen the language they previously selected should be selected.

ZavenArra:
@Elforama I think this PR provides the functionality the @snacks described.

PR #922, opened 2022-03-16
Chat screen

EngrTaofeek:
Corrected few changes in Individual Message List
Made Chat screen


app/src/main/java/org/greenstand/android/TreeTracker/messages/ChatViewModel.kt
```diff
@@ -1,142 +1,76 @@
...
+
+    private val _state = MutableLiveData<ChatState>()
+    val state: LiveData<ChatState> = _state
 
-    fun addMessage(msg: DirectMessage) {
-        _messages.add(0, msg) // Add to the beginning of the list
+    init {
+        viewModelScope.launch {
+            val currentUser = users.getUser(userId)
+            val messages = messagesRepo.getDirectMessages(currentUser!!.wallet, otherChatIdentifier)
```
EngrTaofeek, line 39: @Elforama the messages in state object is always null. Either this call returns null or something. I also try to check the database inspector but the message entity wasn't showing.

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/MessagesRepo.kt
```diff
@@ -56,14 +56,13 @@ class MessagesRepo(
...
-        from: String,
-        to: String): Flow<List<DirectMessage>> {
+        otherChatIdentifier: String): Flow<List<DirectMessage>> {
         return messagesDao.getDirectMessagesForWallet(wallet)
             .map { messages ->
                 messages
                     .map { convertMessageEntityToMessage(it) }
                     .filterIsInstance<DirectMessage>()
-                    .filter { (it.from == from || it.from == to) && (it.to == to || it.to == from) }
+                    .filter { (it.from == wallet || it.from == otherChatIdentifier) && (it.to == otherChatIdentifier || it.to == wallet) }
```
EngrTaofeek, line 65: @Elforama or could the changes here be affecting the query results

app/src/main/java/org/greenstand/android/TreeTracker/messages/ChatViewModel.kt
Elforama: You are not subscribing to the data from the flow.

It should be  
'
messages.repo.getDirectmessages(...)
.collect {
state = state.copy(messages = it, ...)

}
`

app/build.gradle
```diff
@@ -220,6 +220,8 @@ dependencies {
     //Permissions management library for Jetpack Compose
     implementation "com.google.accompanist:accompanist-permissions:0.21.1-beta"
 
+    implementation  "androidx.compose.material3:material3:1.0.0-alpha06"
```
Elforama, line 223: What's this for?

app/src/main/java/org/greenstand/android/TreeTracker/messages/ChatScreen.kt
```diff
@@ -0,0 +1,286 @@
...
+    val scrollState = rememberLazyListState()
+    val scrollBehavior = remember { TopAppBarDefaults.pinnedScrollBehavior() }
+    val state by viewModel.state.observeAsState(ChatState())
+
+
+    Scaffold(
+        topBar = {
+            ActionBar(
+                leftAction = {
+                   DepthButton(onClick = { /*TODO*/ }, modifier =Modifier
```
Elforama, line 80: You can move this button into its own composable.

app/src/main/java/org/greenstand/android/TreeTracker/messages/ChatScreen.kt
```diff
@@ -0,0 +1,286 @@
...
+                .testTag(ConversationTestTag)
+                .fillMaxSize()
+        ) {
+            messages?.map { message ->
+                for (index in message.indices) {
+                    val prevAuthor = message.getOrNull(index - 1)?.from
+                    val nextAuthor = message.getOrNull(index + 1)?.from
+                    val content = message[index]
+                    val isFirstMessageByAuthor = prevAuthor != content.from
+                    val isLastMessageByAuthor = nextAuthor != content.from
```
Elforama, line 205: You can move this logic into the viewModel. Then the viewmodel can expose a custom data class just for this screen such as "ChatMessage" that has everything ready for the view to just render

app/src/main/java/org/greenstand/android/TreeTracker/messages/ChatScreen.kt
```diff
@@ -0,0 +1,286 @@
...
+    msg: DirectMessage,
+    isAdmin: Boolean,
+    isFirstMessageByAuthor: Boolean,
+    isLastMessageByAuthor: Boolean
+) {
+    val spaceBetweenAuthors = if (isLastMessageByAuthor) Modifier.padding(top = 8.dp) else Modifier
+    Row(modifier = spaceBetweenAuthors) {
+        TextMessage(
+            msg = msg,
+            isAdmin = isAdmin,
```
Elforama, line 232: I'd avoid using "admin" in the code. It could be possible later that it's not an admin

app/src/main/java/org/greenstand/android/TreeTracker/messages/ChatViewModel.kt
```diff
@@ -0,0 +1,76 @@
...
+import org.greenstand.android.TreeTracker.models.messages.DirectMessage
+import org.greenstand.android.TreeTracker.models.messages.MessagesRepo
+import org.greenstand.android.TreeTracker.models.user.User
+import org.koin.core.component.KoinComponent
+import org.koin.core.component.get
+import java.util.*
+
+
+data class ChatState(
+    val messages: Flow<List<DirectMessage>>? = null,
```
Elforama, line 21: This should just be a list of messages, not a Flow

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/MessagesRepo.kt
Elforama: This looks fine

app/src/main/java/org/greenstand/android/TreeTracker/view/SelectableItemDetail.kt
```diff
@@ -82,12 +86,29 @@ fun SelectableImageDetail(
...
+                messageTypeText?.let {
+                    Text(
+                        text = it.uppercase(),
+                        color = CustomTheme.textColors.darkText,
+                        fontWeight = FontWeight.Bold,
+                        style = CustomTheme.typography.regular,
+                        textAlign = TextAlign.Center,
+                        modifier = Modifier.padding(top = 110.dp)
+                    )
+                }
```
Elforama, line 111: Let's not put this into the SelectableItem details. This already takes in "content" which is how we show the text below the image. 

At the time, the image was always the same type, but now that it can show different types of images we should pass in a "header" composable that does al that work for us. Then each different type of list item can pass in it's own type of header.

When we start passing in tons of params like this, things get very confusing and don't work properly. For instance, if I passed in a value for each param in this function, it would look broken. The function should be safe to use for all params.

app/src/main/java/org/greenstand/android/TreeTracker/messages/ChatScreen.kt
ZavenArra: yes, don't think of the the incoming messages as from 'admin' but rather from another user.  for the time being it's a manager that will send messages, but we may allow direct person-to-person messaging at some point.

app/src/main/java/org/greenstand/android/TreeTracker/messages/ChatViewModel.kt
EngrTaofeek: Thanks it works now.

app/build.gradle
EngrTaofeek: It is for the nested scroll behaviour in the Chat screen

EngrTaofeek:
@Elforama I have updated with the requested changes. Once this is merged I would proceed to the Announcement Message. I should complete that by Thursday evening.

Elforama (changes requested):
This looks like a good start. Please address all the comments, and feel free to ask questions.

ZavenArra:
I have tested this PR for functionality, and it looks pretty good.  I think we should merge it and allow testers to start trying.

We could file all unresolved conversations as tickets in M2.1 or M2.1.1

app/src/main/java/org/greenstand/android/TreeTracker/messages/ChatScreen.kt
```diff
@@ -0,0 +1,291 @@
...
+                text = message.body,
+                color = CustomTheme.textColors.lightText,
+                fontWeight = FontWeight.Bold,
+                style = CustomTheme.typography.regular,
+            )
+    }
+}
+
+@Composable
+fun OtherChatButton(){
```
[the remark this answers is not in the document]
EngrTaofeek: @Elforama from the sketch file it's a depth button. The sketch file is more accurate than the Figma file as some things were not imported properly from sketch into Figma. See screenshot below.
Screenshot 2022-03-17 at 21 29 37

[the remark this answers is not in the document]
Elforama: I think we should ignore that design choice. It's a button that does nothing. That's a bad UX. 

@ZavenArra what are your thoughts here? I'm suggesting the "admin" and profile picture are just pictures, and are not made to be non functioning buttons

app/src/main/java/org/greenstand/android/TreeTracker/messages/ChatViewModel.kt
```diff
@@ -0,0 +1,92 @@
...
+        }
+    }
+
+    fun updateDraftText(text: String) {
+        _state.value = _state.value!!.copy(
+            draftText = text
+        )
+    }
+
+    fun checkChatAuthor(index :Int,isFirstMessage: Boolean ):Boolean{
```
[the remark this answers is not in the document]
EngrTaofeek: Checks the author of previous message to determine the space between the new message. Usecase : space of 4 dp is used if two consecutive messages have the same author then 8.dp if they are different authors. I could remove it though if it's unnecessary.

app/src/main/java/org/greenstand/android/TreeTracker/messages/ChatScreen.kt
```diff
@@ -0,0 +1,291 @@
...
+        } else {
+            // Between bubbles
+            Spacer(modifier = Modifier.height(4.dp))
+        }
+    }
+}
+
+@Composable
+fun ChatItemBubble(
+    message: DirectMessage,
```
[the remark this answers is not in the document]
EngrTaofeek: @Elforama yes I followed it. It's better this way. For the announcement message where we need links it won't be in the body it would be videoLink if I am correct. So instead of passing different parameters of DirectMessage we would just pass DirectMessage. Another use case is if we want to add time or date to the chat bubble.

app/src/main/java/org/greenstand/android/TreeTracker/messages/ChatScreen.kt
```diff
@@ -0,0 +1,291 @@
...
+                    state = state,
+                    modifier = Modifier.weight(1f),
+                    scrollState = scrollState,
+                    viewModel = viewModel
+                )
+            Box(modifier = Modifier
+                .padding(top = 4.dp, start = 4.dp, end = 4.dp, bottom = 80.dp)
+                .fillMaxWidth()
+                .wrapContentHeight()){
+                TextField(
```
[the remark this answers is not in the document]
EngrTaofeek: I just followed the design. I would add a send button
Screenshot 2022-03-17 at 21 48 57

PR #923, opened 2022-03-16
Add message uploading

Elforama:
tested with direct messages, not surveys

@ZavenArra Check the MessageRequest object for correctness. That is the json being sent. It's used for both messages and survey responses


ZavenArra (approved):
Payload looks good.

PR #928, opened 2022-03-17
feat: enable messages in beta build

ZavenArra:


Elforama:
you'll need to use this flag instead of the debug flag in the  SyncDataUseCase. Then this should be good to go

Issue #888, opened 2022-03-07
First time I press upload, no snack message shows.


ZavenArra:

sebastiangaertner:
Can confirm this

sidsharma2002:
@Elforama can you guide me on this?

sidsharma2002:
@Elforama in the dashboardViewModel, I changed the _isSyncing's initial value from null to false. The issue seems to be solved, is there a specific reason why its init value was null rather than false ? if not I'll raise the Pr

PR #940, opened 2022-03-18
Fix  : No snackbar when sync button is clicked for the first time

sidsharma2002:
#888 Initially the flow was wrong because the starting value of _isSyncing is null. When we open the next time,   unexpectedly the worker emits a work with succeeded state , the null check in the when block prevents the snackbar to show but sets its value as false. Then, when we click the sync button everything happens as intended (due to the false value set).

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/dashboard/DashboardViewModel.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/dashboard/DashboardViewModel.kt
@@ -40,16 +40,12 @@ class DashboardViewModel(
     locationDataCapturer: LocationDataCapturer,
 ) : ViewModel() {
 
-    init {
-        locationDataCapturer.stopGpsUpdates()
-    }
-
     private val _state = MutableLiveData<DashboardState>()
     val state: LiveData<DashboardState> = _state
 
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/dashboard/DashboardViewModel.kt
```diff
@@ -70,60 +66,44 @@ class DashboardViewModel(
...
-                }
-
+                showSnackBar?.invoke(R.string.sync_preparing)
                 _isSyncing = true
             }
         }
     }
 
     init {
-        updateData()
```
Elforama, line 119: This method was here to refresh the screen when going back to it after capturing trees. We still need it.

Elforama (changes requested):
Great work, it looks like the issue is resolved.

There is just one issue. When capturing trees before the user syncs, the dashboard doesn't show the tree count. Adding back the updateData method seems to fix it.

Elforama (approved):
Awesome job, it works well!

Please fix any conflicts and I'll get this merged.

ZavenArra (approved):
Tested and works well.

Issue #946, opened 2022-03-18
 java.lang.IllegalArgumentException: Key d was already used.


ZavenArra:
I got this crash on a user that has 3 surveys.  Clicking any survey and then clicking the next icon crashes the app with the trace below.

Works fine if there is only 1 survey

2022-03-18 15:11:40.251 23712-23712/org.greenstand.android.TreeTracker.dev E/AndroidRuntime: FATAL EXCEPTION: main
    Process: org.greenstand.android.TreeTracker.dev, PID: 23712
    java.lang.IllegalArgumentException: Key d was already used. If you are using LazyColumn/Row please make sure you provide a unique key for each item.
        at androidx.compose.ui.layout.SubcomposeLayoutState.subcompose$ui_release(SubcomposeLayout.kt:214)
        at androidx.compose.ui.layout.SubcomposeLayoutState$Scope.subcompose(SubcomposeLayout.kt:490)
        at androidx.compose.foundation.lazy.layout.LazyLayoutPlaceablesProvider.getAndMeasure-0kLqBqw(LazyMeasurePolicy.kt:58)
        at androidx.compose.foundation.lazy.LazyMeasuredItemProvider.getAndMeasure-ZjPyQlc(LazyMeasuredItemProvider.kt:45)
        at androidx.compose.foundation.lazy.LazyListMeasureKt.measureLazyList-DCyOChk(LazyListMeasure.kt:140)
        at androidx.compose.foundation.lazy.LazyListKt$rememberLazyListMeasurePolicy$1$1.measure-3p2s80s(LazyList.kt:230)
        at androidx.compose.foundation.lazy.layout.LazyLayoutKt$LazyLayout$2.invoke-0kLqBqw(LazyLayout.kt:55)
        at androidx.compose.foundation.lazy.layout.LazyLayoutKt$LazyLayout$2.invoke(LazyLayout.kt:44)
        at androidx.compose.ui.layout.SubcomposeLayoutState$createMeasurePolicy$1.measure-3p2s80s(SubcomposeLayout.kt:355)
        at androidx.compose.ui.node.InnerPlaceable.measure-BRTryo0(InnerPlaceable.kt:55)
        at androidx.compose.foundation.layout.PaddingValuesModifier.measure-3p2s80s(Padding.kt:417)
        at androidx.compose.ui.node.ModifiedLayoutNode.measure-BRTryo0(ModifiedLayoutNode.kt:39)
[...]

PR #950, opened 2022-03-19
fix survey crash due to un unique key

Elforama:
fixes https://github.com/Greenstand/treetracker-android/issues/946


ZavenArra:
tested with many surveys, works well.

Issue #970, opened 2022-04-05
Visualize convergence while capturing


ZavenArra:
When waiting for convergence while capturing an image of tree, we want to show a graphic that indicated how convergence is progressing toward an acceptable level, and indicate that the camera must be held still.

A graphic that is always shown indicating to hold the camera still is probably necessary.

Additionally, use the following match to display improvement in convergence until 100% converged

percent converged = min(100, 100 * convergence_target_value /  current_convergence_value)

EngrTaofeek:
@ZavenArra I think the instruction on holding the camera still should also be included in the tutorials also. I would work on this issue

ZavenArra:
OK great

EngrTaofeek:
![](https://user-images.githubusercontent.com/38661872/165352881-1085eff2-f13a-45d3-a0f6-284246efb0aa.jpeg)
![](https://user-images.githubusercontent.com/38661872/165352891-3fbf6a57-5e28-4ed4-aca1-5cc6b0fe42a6.jpeg)
@ZavenArra and @sebastiangaertner  this what this PR does adds a dark background to the progress bar and shows the percentage converged along with instruction once the capture button is clicked.

ZavenArra:
I would like to release this PR and make it available for testers and use that to drive feedback.  @Elforama what do you think about adopting a process along those lines?

EngrTaofeek:
fixed in #972

PR #972, opened 2022-04-13
Determinate progress bar

EngrTaofeek:
Created determinate progress bar to show progress based on percentage of convergence. Also added partially transparent dark background to make the progress bar more visible while tracking. Then added tracking instructions indicating phone has to be still while tracking.
One minor challenge(would be highlighted in review) to be fixed before it's ready to be merged.

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureViewModel.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureViewModel.kt
@@ -23,7 +23,7 @@ data class TreeCaptureState(
     val isCreatingFakeTrees: Boolean = false,
     val isLocationAvailable: Boolean? = null,
     val showCaptureTutorial: Boolean? = null,
-    val convergencePercentage: Float? = null
+    val convergencePercentage: Float = 0.0f,
 )
 
 class TreeCaptureViewModel(
@@ -45,8 +45,7 @@ class TreeCaptureViewModel(
     }
 
     suspend fun captureLocation() {
-        _state.value = _state.value?.copy(isGettingLocation = true)
-//        _state.value = _state.value?.copy(convergencePercentage = treeCapturer.getConvergencePercentage())
+        _state.value = _state.value?.copy(isGettingLocation = true, convergencePercentage = locationDataCapturer.percentageConvergence)
         _state.value = _state.value?.copy(isLocationAvailable = treeCapturer.pinLocation(), isGettingLocation = false)
     }
 
--- a/app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureViewModel.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureViewModel.kt
@@ -23,6 +23,7 @@ data class TreeCaptureState(
     val isCreatingFakeTrees: Boolean = false,
     val isLocationAvailable: Boolean? = null,
     val showCaptureTutorial: Boolean? = null,
+    val convergencePercentage: Float? = null
 )
 
 class TreeCaptureViewModel(
@@ -45,6 +46,7 @@ class TreeCaptureViewModel(
 
     suspend fun captureLocation() {
         _state.value = _state.value?.copy(isGettingLocation = true)
+//        _state.value = _state.value?.copy(convergencePercentage = treeCapturer.getConvergencePercentage())
         _state.value = _state.value?.copy(isLocationAvailable = treeCapturer.pinLocation(), isGettingLocation = false)
     }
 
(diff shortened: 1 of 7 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureViewModel.kt
```diff
@@ -45,8 +45,7 @@ class TreeCaptureViewModel(
     }
 
     suspend fun captureLocation() {
-        _state.value = _state.value?.copy(isGettingLocation = true)
-//        _state.value = _state.value?.copy(convergencePercentage = treeCapturer.getConvergencePercentage())
+        _state.value = _state.value?.copy(isGettingLocation = true, convergencePercentage = locationDataCapturer.percentageConvergence)
```
EngrTaofeek, line 48: @Elforama @ZavenArra this is the only issue left with this PR. The locationDataCapturer.percentageConvergence returns only the default value and doesn't update the state when tracking. But from the logs I can see the value been updated. Any tip on how to make this work

Elforama: Where are you putting the logs that it gets updated.

app/src/main/java/org/greenstand/android/TreeTracker/models/location/LocationDataCapturer.kt
```diff
@@ -75,12 +75,12 @@ class LocationDataCapturer(
...
                     if (longStdDev != null && latStdDev != null) {
-//                        val minimumConvergenceRatio = min(locationDataConfig.latStdDevThreshold.div(latStdDev).toFloat(),locationDataConfig.lonStdDevThreshold.div(longStdDev).toFloat())
-//                        percentageConvergence = min(1f,minimumConvergenceRatio)
-//                        Timber.d(
-//                            "Percentage convergence " +
-//                                    "[${percentageConvergence}]. \n "
-//                        )
+                        val minimumConvergenceRatio = min(locationDataConfig.latStdDevThreshold.div(latStdDev).toFloat(),locationDataConfig.lonStdDevThreshold.div(longStdDev).toFloat())
+                        percentageConvergence = min(1f,minimumConvergenceRatio)
+                        Timber.d(
```
EngrTaofeek, line 80: @Elforama this is where I logged the value which was been updated as the capture converges

app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureViewModel.kt
Elforama: You're just reading a float, to get the value updated you need to listen to it. You'll need to add a convergence listener to the class that updates the value, then you can listen to the value to update the state.

app/src/main/java/org/greenstand/android/TreeTracker/models/location/LocationDataCapturer.kt
```diff
@@ -36,8 +37,7 @@ class LocationDataCapturer(
         private set
     private var convergenceStatus: ConvergenceStatus? = null
     private var areLocationUpdatesOn: Boolean = false
-    var percentageConvergence: Float? = null
-        private set
+    var percentageConvergence : Float = 0f
```
EngrTaofeek, line 40: @Elforama this

EngrTaofeek:
@ZavenArra @sebastiangaertner this what this PR does, it adds a dark background to the progress bar and shows the percentage converged along with instruction once the capture button is clicked.

Elforama:
A few bit of feedback on this:
- The progress bar and progress are not aligned
- We don't need the decimal place for the percentage
- Do we want to be blocking the users view as they try taking a picture? Perhaps another type of UI would work better, such as a progress line that goes around the edge of the screen?

app/src/main/java/org/greenstand/android/TreeTracker/models/TreeCapturer.kt
```diff
@@ -5,6 +5,7 @@ import org.greenstand.android.TreeTracker.models.location.LocationDataCapturer
 import java.io.File
 import java.util.UUID
 import org.greenstand.android.TreeTracker.usecases.CreateTreeUseCase
+import kotlin.math.min
```
Elforama, line 8: No need for this

Elforama (changes requested):

EngrTaofeek:
> A few bit of feedback on this:
> 
> * The progress bar and progress are not aligned
> * We don't need the decimal place for the percentage
> * Do we want to be blocking the users view as they try taking a picture? Perhaps another type of UI would work better, such as a progress line that goes around the edge of the screen?

@Elforama I addressed the first and second in a new commit. For the third, there were complaints that the spinning progress bar was not visible enough which is why I added a dark background to make it more visible under sun light. There were also complaints farmers do not wait for tree convergence to complete before leaving the front of the tree being captured which results into saving random photos that's why the instruction is always there in the dialog while converging.

EngrTaofeek:
fixes #970

ZavenArra:
image

Let's make this one final fix, and then I think this PR is good to merge.

EngrTaofeek:
updated fix 
image

ZavenArra (approved):
Sending this to QC

PR #975, opened 2022-05-06
Fixed location enabled update

### LINKED ISSUE #974: Status of location only in initial capture and not subsequent ones.
When location is turned off before capturing tree and the user is directed to put it on, the state of location is not updated/observed and keeps redirecting the planter in subsequent captures even when the location is on.

EngrTaofeek:
fixes #974  Modified logic to check if location is turned on everytime after checking for location permissions.


EngrTaofeek:
This updates the app to check if location is enabled on every tree capture. Previously it was only checked on the initial capture and keeps redirecting the user back to turn on Locations. It is only apparent on physical device and not real emulator.

EngrTaofeek:
Merged in another PR #1006

PR #979, opened 2022-06-01
Fixed validation issues from actionBar

### LINKED ISSUE #977: VALIDATION FAILS FROM ACTION BUTTON original issue "Depth Button Scope Bug #984"
After email/phone input credentials are validated properly when clicked on done button from keyboard but shows error when clicking action right arrow button. Validation and error message should work both from the keyboard and action button.

EngrTaofeek:
Discovered a strange issue. The value of state in the action bar(bottom bar in the scaffold) always used the default sign up state which is different from the values in the child/body of scaffold. So I refactored the state items and ViewModels as fields and instantiated them in the child/body of the scaffold. So the validation from action always use the current state. fixes #977

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/signup/CredentialEntryView.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/signup/CredentialEntryView.kt
@@ -57,6 +57,10 @@ import org.greenstand.android.TreeTracker.view.DepthButton
 import org.greenstand.android.TreeTracker.view.LanguageButton
 import org.greenstand.android.TreeTracker.view.TopBarTitle
 import org.greenstand.android.TreeTracker.view.UserButton
+
+private lateinit var scaffoldState: SignUpState
+private lateinit var actionBarViewModel: SignupViewModel
+
 @Composable
 fun CredentialEntryView(viewModel: SignupViewModel, state: SignUpState) {
     val navController = LocalNavHostController.current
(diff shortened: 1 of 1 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/signup/CredentialEntryView.kt
```diff
@@ -57,6 +57,10 @@ import org.greenstand.android.TreeTracker.view.DepthButton
 import org.greenstand.android.TreeTracker.view.LanguageButton
 import org.greenstand.android.TreeTracker.view.TopBarTitle
 import org.greenstand.android.TreeTracker.view.UserButton
+
+private lateinit var scaffoldState: SignUpState
+private lateinit var actionBarViewModel: SignupViewModel
```
Elforama, line 62: We shouldn't have global state like this. It makes it hard to debug/reason about the code and will most likely introduce bugs now or later.

Elforama (changes requested):
Global state like this is an anti pattern. Please find another way to resolve this.

app/src/main/java/org/greenstand/android/TreeTracker/signup/CredentialEntryView.kt
Elforama: We need to find a different way to fix this. This may work in the short run, but it's basically asking for complicated bugs in the future. I also think there are probably edge cases here that have not been tested. With a global viewmodel, one signup input can persist when going back to the screen.

EngrTaofeek:
@Elforama #984 was the actual issue.

EngrTaofeek:
fixed in #1000

PR #981, opened 2022-06-07
Display message notifications

### LINKED ISSUE #971: Upload messages if no tree captured

### LINKED ISSUE #980: Show notification symbol
Message Button on dashboard and IndividualMessageListScreen items should show notification symbol when there are unread messages.

EngrTaofeek:
Fixes #980 Fixes #971 
-Syncs messages from dashboard without uploading tree(clicking message button)
-Updates dashboard UI to show notification for unread messages and also dismiss them after they are read
- Update UserEntity to observe unread messages per user and also display notification for unread messages per user.
- Minor code refactor

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/database/AppDatabase.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/database/AppDatabase.kt
@@ -18,7 +18,7 @@ import org.greenstand.android.TreeTracker.database.legacy.entity.TreeAttributeEn
 import org.greenstand.android.TreeTracker.database.legacy.entity.TreeCaptureEntity
 
 @Database(
-    version = 9,
+    version = 8,
     exportSchema = true,
     entities = [
         PlanterCheckInEntity::class,
@@ -33,7 +33,7 @@ import org.greenstand.android.TreeTracker.database.legacy.entity.TreeCaptureEnti
         DeviceConfigEntity::class,
     ],
     autoMigrations = [
-        AutoMigration(from = 8, to = 9)
+        AutoMigration(from = 7, to = 8)
     ],
 )
 @TypeConverters(Converters::class)
@@ -58,7 +58,6 @@ abstract class AppDatabase : RoomDatabase() {
                             MIGRATION_4_5,
                             MIGRATION_5_6,
                             MIGRATION_6_7,
-                            MIGRATION_8_9
                         )
                         .build()
                 }
--- a/app/src/main/java/org/greenstand/android/TreeTracker/database/AppDatabase.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/database/AppDatabase.kt
@@ -18,7 +18,7 @@ import org.greenstand.android.TreeTracker.database.legacy.entity.TreeAttributeEn
 import org.greenstand.android.TreeTracker.database.legacy.entity.TreeCaptureEntity
 
 @Database(
-    version = 8,
+    version = 9,
     exportSchema = true,
     entities = [
         PlanterCheckInEntity::class,
@@ -33,7 +33,7 @@ import org.greenstand.android.TreeTracker.database.legacy.entity.TreeCaptureEnti
         DeviceConfigEntity::class,
     ],
     autoMigrations = [
-        AutoMigration(from = 7, to = 8)
+        AutoMigration(from = 8, to = 9)
     ],
 )
 @TypeConverters(Converters::class)
@@ -58,6 +58,7 @@ abstract class AppDatabase : RoomDatabase() {
                             MIGRATION_4_5,
                             MIGRATION_5_6,
                             MIGRATION_6_7,
+                            MIGRATION_8_9
                         )
                         .build()
                 }
--- a/app/src/main/java/org/greenstand/android/TreeTracker/database/Migrations.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/database/Migrations.kt
@@ -46,6 +46,15 @@ val MIGRATION_4_5 = object : Migration(4, 5) {
     }
 }
 
+val MIGRATION_8_9 = object : Migration(8, 9) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        val sql = """ALTER TABLE `user`
+            | ADD COLUMN `unread_messages_available` INTEGER NOT NULL DEFAULT 0
+            | """.trimMargin()
+        database.execSQL(sql)
+    }
+}
+
 val MIGRATION_3_4 = object : Migration(3, 4) {
 
     override fun migrate(database: SupportSQLiteDatabase) {
@@ -69,7 +78,7 @@ val MIGRATION_3_4 = object : Migration(3, 4) {
                 |  `bundle_id` TEXT
                 |  )""".trimMargin()
             database.execSQL(createPlanterInfo)
-            val createIndexPlanterIdentifier = """
+            val createIndexPlanterIdentifier = """ 
                 | CREATE INDEX IF NOT EXISTS 
                 | `index_planter_identifier` ON `planter_info` (`planter_identifier`)
                 | """.trimMargin()
--- a/app/src/main/java/org/greenstand/android/TreeTracker/database/Migrations.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/database/Migrations.kt
@@ -46,15 +46,6 @@ val MIGRATION_4_5 = object : Migration(4, 5) {
     }
 }
 
-val MIGRATION_8_9 = object : Migration(8, 9) {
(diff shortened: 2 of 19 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/database/Migrations.kt
```diff
@@ -46,6 +46,15 @@ val MIGRATION_4_5 = object : Migration(4, 5) {
...
 
+val MIGRATION_8_9 = object : Migration(8, 9) {
+    override fun migrate(database: SupportSQLiteDatabase) {
+        val sql = """ALTER TABLE `user`
+            | ADD COLUMN `unread_messages_available` INTEGER NOT NULL DEFAULT 0
+            | """.trimMargin()
+        database.execSQL(sql)
+    }
+}
+
```
Elforama, line 57: No need for this if we are using the automigration

app/src/main/java/org/greenstand/android/TreeTracker/database/AppDatabase.kt
```diff
@@ -58,6 +58,7 @@ abstract class AppDatabase : RoomDatabase() {
                             MIGRATION_4_5,
                             MIGRATION_5_6,
                             MIGRATION_6_7,
+                            MIGRATION_8_9
```
Elforama, line 61: Not needed, since we're using the auto migration

app/src/main/java/org/greenstand/android/TreeTracker/messages/announcementmessage/AnnouncementViewModel.kt
```diff
@@ -40,6 +40,7 @@ class AnnouncementViewModel(
                 currentUrl = announcement.videoLink
             )
             messagesRepo.markMessageAsRead(messageId)
+            messagesRepo.syncMessages()
```
Elforama, line 43: This code mainly marks the message as read. There shouldn't be anything to sync with the server yet. Can you explain why we're syncing here?

app/src/main/java/org/greenstand/android/TreeTracker/messages/directmessages/ChatViewModel.kt
```diff
@@ -45,6 +45,7 @@ class ChatViewModel(
                 )
                 val unreadMessages = messages.filterNot { it.isRead }.map { it.id }
                 messagesRepo.markMessagesAsRead(unreadMessages)
+                messagesRepo.syncMessages()
```
Elforama, line 48: Same as before, why are we syncing here?

app/src/main/java/org/greenstand/android/TreeTracker/messages/survey/SurveyViewModel.kt
```diff
@@ -42,6 +42,7 @@ class SurveyViewModel(
                 currentQuestion = survey.questions[currentQuestionIndex]
             )
             messagesRepo.markMessageAsRead(messageId)
+            messagesRepo.syncMessages()
```
Elforama, line 45: Why sync?

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/MessagesRepo.kt
```diff
@@ -111,6 +112,14 @@ class MessagesRepo(
         return convertMessageEntityToMessage(messagesDao.getMessage(id)!!) as SurveyMessage
     }
 
+    suspend fun checkForUnreadMessages(): Boolean{
+        return messagesDao.getTotalUnreadMessagesCount() >= 1
+    }
+
+    suspend fun checkForUnreadMessagesPerUser(wallet: String): Boolean{
```
Elforama, line 119: This returns a boolean. Let's name it something like "doesWalletHaveUnreadMessages(wallet)"

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/database/MessagesDAO.kt
```diff
@@ -58,6 +58,12 @@ interface MessagesDAO {
     @Query("UPDATE messages SET is_read = 1 WHERE id IN (:id)")
     suspend fun markMessageAsRead(id: List<String>)
 
+    @Query("SELECT COUNT(*) FROM messages WHERE is_read = 0")
+    suspend fun getTotalUnreadMessagesCount(): Int
+
+    @Query("SELECT COUNT(*) FROM messages WHERE wallet = (:wallet) AND is_read = 0")
+    suspend fun getTotalUnreadMessagesPerUserCount(wallet: String): Int
```
Elforama, line 65: This name seems longer than it needs to be. Can we try something like 'getUnreadMessageCountForWallet'?

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/database/MessagesDAO.kt
```diff
@@ -58,6 +58,12 @@ interface MessagesDAO {
     @Query("UPDATE messages SET is_read = 1 WHERE id IN (:id)")
     suspend fun markMessageAsRead(id: List<String>)
 
+    @Query("SELECT COUNT(*) FROM messages WHERE is_read = 0")
+    suspend fun getTotalUnreadMessagesCount(): Int
```
Elforama, line 62: we can shorten to 'getUnreadMessageCount'

app/src/main/java/org/greenstand/android/TreeTracker/database/entity/UserEntity.kt
```diff
@@ -36,6 +36,9 @@ data class UserEntity(
     @ColumnInfo(name =
     "power_user", defaultValue = "0")
     var powerUser: Boolean,
+    @ColumnInfo(name =
+    "unread_messages_available", defaultValue = "0")
+    var unreadMessagesAvailable: Boolean,
```
Elforama, line 41: I think this makes the code more complicated than it needs to be. We know which messages are read, and which user they belong to. We can easily query how many messages a user has with that info without keeping track in the user object. 
When you place this info in the user object, it's more data that has to now keep in sync.

Elforama (changes requested):
I think we can simplify this code. Please check the comments.

app/src/main/java/org/greenstand/android/TreeTracker/database/AppDatabase.kt
EngrTaofeek: It was crashing without adding this

app/src/main/java/org/greenstand/android/TreeTracker/database/entity/UserEntity.kt
EngrTaofeek: The user object updates this screen 
Screenshot_20220612_133712
so it is easier to have it as a value in the User object which updates each user button on the screen.

Elforama: I agree, so we can add it to the user object, but not the user entity. This is why we have a repo layer, we can separate how we store things with what the rest of the app has access to. We can query the messages when returning a user object to fill that fiels.

EngrTaofeek: @Elforama sorry I was away for a while.  Yeah this makes more sense actually.

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/MessagesRepo.kt
Elforama: Don't forget to make this change.

app/src/main/java/org/greenstand/android/TreeTracker/models/user/Users.kt
```diff
@@ -36,6 +38,14 @@ class Users(
         return createUser(dao.getUserByWallet(wallet))
     }
 
+    suspend fun checkForUnreadMessagesPerUser(wallet: String): Boolean{
+        return messagesDao.getUnreadMessageCountForWallet(wallet) >= 1
+    }
```
Elforama, line 43: You're duplicating the logic that's already in the messages repo. Just use the messagesRepo.

Adding the messagesrepo to user will give a circular dependency error. To solve this, make a new class that syncs messages and move the sync message method from the messagesRepo into that.

Elforama (changes requested):
Left some comments

app/src/main/java/org/greenstand/android/TreeTracker/models/user/Users.kt
EngrTaofeek: This is no longer used in the message repo I forgot to delete it in the previous commit.  
Going by the solution you suggested I would just be refactoring the sync logic in the message repo to another class it still caused the same circular dependency error.

Issue #984, opened 2022-07-02
Depth Button Scope Bug


EngrTaofeek:
The depthButton always uses a new scope while clicked regardless. This actually occurs in detectTap gestures, Line 416 of https://github.com/Greenstand/treetracker-android/blob/master/app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt So Tried running the events in clickable which used the right scope but didn't give the desired animation effect. I tested with other components like button, box etc same thing scopes remained the same in clickable but uses default in detectGesture.

This is why we this issue persisted #977 which was why this https://github.com/Greenstand/treetracker-android/pull/844 also was not completed.

EngrTaofeek:
fixed in #1000

PR #999, opened 2022-08-04
modified design and added log to show the scope bug

EngrTaofeek:
modified design and added log to show the scope bug. Clickable events can be handled in Modifier.clickable{} or Modifier.pointerInput(true) { detectTapGestures(... To describe the bug I created two buttons that handle events in both ways and logs the current value. For the first one it shows the current value but for the second approach it keeps using the default value. The onClick events in the depth buttons are handled in the second approach which is why the arrow button fails to validate except by keyboard and other issues discussed here #984


Elforama:
Thanks for giving this detailed example.

PR #1002, opened 2022-08-10
added info button

SaraGupt:
Screenshot_20220811-153623

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
@@ -10,6 +10,7 @@ import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.material.AlertDialog
 import androidx.compose.material.ButtonColors
 import androidx.compose.material.Snackbar
+import androidx.compose.material.SnackbarDefaults.backgroundColor
 import androidx.compose.material.SnackbarHost
 import androidx.compose.material.SnackbarHostState
 import androidx.compose.material.Text
@@ -30,11 +31,13 @@ import androidx.compose.ui.draw.clip
 import androidx.compose.ui.geometry.CornerRadius
 import androidx.compose.ui.geometry.Offset
 import androidx.compose.ui.geometry.Size
+import androidx.compose.ui.graphics.Brush
 import androidx.compose.ui.graphics.Color
 import androidx.compose.ui.graphics.drawscope.Stroke
 import androidx.compose.ui.graphics.painter.Painter
 import androidx.compose.ui.input.pointer.pointerInput
 import androidx.compose.ui.layout.ContentScale
+import androidx.compose.ui.modifier.modifierLocalConsumer
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.text.font.FontWeight
 import androidx.compose.ui.tooling.preview.Preview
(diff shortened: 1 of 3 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -396,7 +451,7 @@ fun DepthButton(
     content: @Composable (BoxScope.() -> Unit),
 ) {
     val contentColor by colors.contentColor(isEnabled)
-    val onClickState = rememberUpdatedState(onClick)
```
EngrTaofeek, line 399: Please reverse this. This is actually a fix to one terrible bug.

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -415,7 +470,7 @@ fun DepthButton(
                     if (!isEnabled) return@pointerInput
                     detectTapGestures(
                         onTap = {
-                            onClickState.value.invoke()
+                            onClick()
```
EngrTaofeek, line 473: reverse this also and leave the previous line. This change would cause the depthButton to always use default state and won't recreate it. For more info see #1000 and #984

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -121,6 +124,58 @@ fun ApprovalButton(
     }
 }
 
+@Composable
+fun infoButton(modifier: Modifier = Modifier, isEnabled: Boolean,
```
EngrTaofeek, line 128: Kindly rename this to InfoButton. For composable we want to have the name of the composable functions start with a capital letter while other non-composable functions can start with a small letter.

EngrTaofeek (changes requested):
Great work Sara. Kindly take a look at the requested changes and attach screenshot or screen recorded video of the app displaying the change you made.

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -121,6 +124,58 @@ fun ApprovalButton(
...
+        onClick = onClick,
+        shape = DepthSurfaceShape.Circle,
+        depth = 10f
+    ) {
+        infoButtonCircle(
+            modifier = Modifier
+                .size(60.dp),
+            color = AppColors.LightGray,
+            shadowColor = AppColors.GrayShadow,
+        )
```
Elforama, line 144: Instead of recreating the icon, we can reuse R.drawable.info_icon instead. We can reuse a lot of what this button used to be, and place is here instead.

Elforama (changes requested):
Thanks for working on this Sara. There are a few changes that need to be made before it's merged, but this looks good overall.

SaraGupt:
Thanks for the review. I have addressed the comments.

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -121,6 +121,25 @@ fun ApprovalButton(
     }
 }
 
+@Composable
+fun InfoButton(modifier: Modifier = Modifier, isEnabled: Boolean,
```
Elforama, line 125: Since this button is always enabled we don't need to expose "isEnabled" and can hardcode the value for the depth button.

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -121,6 +121,25 @@ fun ApprovalButton(
...
+               onClick: () -> Unit,shape: DepthSurfaceShape = DepthSurfaceShape.Circle) {
```
Elforama, line 126: Please make sure to use proper spacing throughout the code. You can refer to the rest of the code to see how we use spacing as well as the placement of curly braces. Android studio should auto format code for you with a certain button.

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -121,6 +121,25 @@ fun ApprovalButton(
...
+        modifier = modifier.size(70.dp),
+        isEnabled = isEnabled,
+        colors = AppButtonColors.Default,
+        onClick = onClick,
+        shape = DepthSurfaceShape.Circle,
+        depth = 10f
+    )
+    {
+        Image(painter = painterResource(id = R.drawable.info_icon), modifier = Modifier
+            .size(80.dp),contentDescription =null )
```
Elforama, line 138: This shows a white image on a dark button. We should customize the color of the button to match it. The top button color should be the same white color, and the edge "shadow" of the button should be slightly darker.

Elforama (changes requested):
Thanks for making the changes. It's def looking better, but there are a few more changes I'd like you to make that I've shared.

SaraGupt:
Thank You. I have addressed the new comments.

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
SaraGupt: Addressed.! Thank you

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
SaraGupt: Reverted. Thanks!

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
SaraGupt: Thanks for highlight. Addressed.!

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
SaraGupt: Thanks! Addressed.

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
SaraGupt: Addressed..!

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
SaraGupt: Thanks for highlight.! Addressed.

app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
SaraGupt: Addressed..!!

Elforama (approved):

PR #1004, opened 2022-08-11
Updated Org Picker screen

EngrTaofeek:
Updated Org Picker screen

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerScreen.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerScreen.kt
@@ -72,7 +72,8 @@ fun OrgPickerScreen(viewModel: OrgPickerViewModel = viewModel(factory = LocalVie
         LazyVerticalGrid(
             cells = GridCells.Fixed(2),
             modifier = Modifier.fillMaxSize(),
-            horizontalArrangement = Arrangement.SpaceEvenly
+            horizontalArrangement = Arrangement.SpaceEvenly,
+            verticalArrangement = Arrangement.Center
         ) {
             items(state.orgs) { org ->
                 OrgItem(
--- a/app/src/main/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerScreen.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerScreen.kt
@@ -1,16 +1,14 @@
 package org.greenstand.android.TreeTracker.orgpicker
 
-import androidx.compose.foundation.background
-import androidx.compose.foundation.clickable
+import androidx.compose.foundation.ExperimentalFoundationApi
 import androidx.compose.foundation.layout.Arrangement
 import androidx.compose.foundation.layout.Box
-import androidx.compose.foundation.layout.Column
-import androidx.compose.foundation.layout.Spacer
 import androidx.compose.foundation.layout.fillMaxSize
-import androidx.compose.foundation.layout.height
+import androidx.compose.foundation.layout.fillMaxWidth
 import androidx.compose.foundation.layout.padding
 import androidx.compose.foundation.layout.size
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerScreen.kt
```diff
@@ -37,9 +41,21 @@ fun OrgPickerScreen(viewModel: OrgPickerViewModel = viewModel(factory = LocalVie
 
     Scaffold(
         topBar = {
-            ActionBar(
-                centerAction = { TopBarTitle() }
-            )
```
Elforama, line 42: I think it would be nice to keep this to keep the design constant with the dashboard and language select screens. None of them have a title on top, they use a logo.

app/src/main/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerScreen.kt
```diff
@@ -53,17 +69,11 @@ fun OrgPickerScreen(viewModel: OrgPickerViewModel = viewModel(factory = LocalVie
             )
         },
     ) {
-        LazyColumn(
+        LazyVerticalGrid(
+            cells = GridCells.Fixed(2),
             modifier = Modifier.fillMaxSize(),
-            horizontalAlignment = Alignment.CenterHorizontally,
-            verticalArrangement = Arrangement.Center
+            horizontalArrangement = Arrangement.SpaceEvenly
```
Elforama, line 75: Can we please center align the list items vertically?

app/src/main/res/values/strings.xml
```diff
@@ -175,4 +175,5 @@ Privacy Policy
     <string name="tracking_progress_message">Do not move the phone until tracking is completed</string>
     <string name="privacy_policy">Privacy Policy</string>
     <string name="select_tree_height_colour">Select tree height colour</string>
+    <string name="select_organization">Select organization</string>
```
Elforama, line 178: I don't think we need this if we just show a logo. Or perhaps we show it below the logo.

Elforama (changes requested):
Looks good, just a few requested changes.

app/src/main/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerScreen.kt
EngrTaofeek: Since this screen is going to show different organizations available at the same time it's better we leave the header independent of an organisation.

EngrTaofeek: Alternatively we could just use Greenstand logo permanently.

app/src/main/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerScreen.kt
EngrTaofeek: We could change the design to this also. Should it still be grid or just single Columns?

app/src/main/res/values/strings.xml
EngrTaofeek: Yeah under Greenstand logo which would be permanent there regardless of selected organisation.

app/src/main/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerScreen.kt
Elforama: I think leaving as text makes sense in this case.

app/src/main/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerScreen.kt
Elforama: A grid is still good, but since most cases this will be only 2 items, it will look nicer if they are centered.

Elforama (approved):

PR #1009, opened 2022-08-23
Message notifications and display.

EngrTaofeek:
-Sync messages when message button is clicked
-Display unread messages symbol on DashboardScreen.kt
-Display unread messages symbol on UserSelect(Message)


EngrTaofeek:
Addresses all the changes previously requested in #981

PR #1011, opened 2022-08-30
932_Launch SnackBar when Survey completed

SaraGupt:


EngrTaofeek:
@SaraGupt I think you tried using the implementation of Snackbars as found in the dashboard screen which is inappropriate in this case. In the Dashboard the Snackbars where been launched from Workmanager states and that's not the case here. To achieve your aim here you can follow the sample implementation of Snackbar in the CredentialEntryView that should work.

EngrTaofeek:
Also try to rebase your branch.

PR #1021, opened 2022-10-01
Code health and Performance improvements

sidsharma2002:
Changes done in this pr : 
 1. removed unused files :  AdminLoginFragment, ConfigFragment.
 2. suspend function now use specific dispatcher instead of taking that from launched coroutine.
 3. sequential saveMessage jobs will now run in parallel.


Elforama:
Have you tested this with messages in the app?

I think adding a unit test to this would be great. That way we can verify no functionality has chan6

sidsharma2002:
Sure I will send a pr separately for unit tests. Also for my profile I can see only one message right now, I need to think of a way to test on my device this thoroughly.

EngrTaofeek:
@sidsharma2002 just to add to properly test the messaging on your device create a wallet handle(email) with any of the following "handle3@test","handle@test" or "mobile1@test" those email id have pre-populated messages and should sync properly if it doesn't break. Another think like @Elforama rightly said is to write unit test.

sidsharma2002:
Hi @EngrTaofeek I cannot find messages on any of these email ids. I was using this on app from playstore, can you recheck at your end if these ids are correct?

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/MessagesRepo.kt
```diff
@@ -146,24 +152,46 @@ class MessagesRepo(
...
+
+        val saveMessageResponseJobs = mutableListOf<Job>()
+
+        result.messages.forEach {
+            ensureActive()
+            // launch all jobs in parallel and add these to a list.
+            saveMessageResponseJobs.add(launch {
+                saveMessageResponse(wallet, it.copy())
+            })
+        }
```
Elforama, line 165: I'm not too familiar with using jobs. What is the difference between doing this and using async such as:
messages.map { async { saveMessage() } }.forEach{ it.await() }

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/MessagesRepo.kt
```diff
@@ -146,24 +152,46 @@ class MessagesRepo(
...
                 wallet = wallet,
                 lastSyncTime = lastSyncTime,
                 offset = query.offset,
                 limit = query.limit,
             )
-            result.messages.forEach { saveMessageResponse(wallet, it) }
+
+            saveMessageResponseJobs.clear()
+
+            result.messages.forEach {
```
Elforama, line 182: These are small/fast DB transactions. I'm wondering if it's worth adding complexity for a most likely non noticeable performance improvement.

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/MessagesRepo.kt
```diff
@@ -146,24 +152,46 @@ class MessagesRepo(
...
+                ensureActive()
```
Elforama, line 183: We probably don't want to cancel this remaining work if the coroutine is canceled. We used the users network data to download this info, we should save it.

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/MessagesRepo.kt
```diff
@@ -146,24 +152,46 @@ class MessagesRepo(
...
+                saveMessageResponseJobs.add(launch {
+                    saveMessageResponse(wallet, it)
+                })
```
Elforama, line 186: Would using async here make more sense?
saveMessageResponseJobs += async { saveMessageResponse(wallet, it) }

Elforama (changes requested):
Thanks for improving the code base!

I left a few comments. The main question I have is if we should add the complexity to the DB transactions if we don't gain much from it. For a network requests parallel execution can save a ton of time, but for small DB transactions I'm not sure if it's worth it unless we can keep it simple/clean, perhaps with a function that simplifies parallel execution.

runInParrallel(dataInput) { myWork(it) }.awaitCompletion()

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/MessagesRepo.kt
sidsharma2002: they are the same thing in this case. await is for when we have to get a result back from a method eg: fun saveMessages(message) : Boolean but here the method doesn't return anything hence we use launch.

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/MessagesRepo.kt
sidsharma2002: makes sense, will remove it

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/MessagesRepo.kt
sidsharma2002: https://stackoverflow.com/questions/46226518/what-is-the-difference-between-launch-join-and-async-await-in-kotlin-coroutines here's a sof question, the saveMessageResponse is a fire and forget task hence launch is used here.

sidsharma2002:
> Thanks for improving the code base!
> 
> I left a few comments. The main question I have is if we should add the complexity to the DB transactions if we don't gain much from it. For a network requests parallel execution can save a ton of time, but for small DB transactions I'm not sure if it's worth it unless we can keep it simple/clean, perhaps with a function that simplifies parallel execution.
> 
> runInParrallel(dataInput) { myWork(it) }.awaitCompletion()

app/src/main/java/org/greenstand/android/TreeTracker/models/messages/MessagesRepo.kt
Elforama: Let's use async then, it's used for parallel work and removes the need to maintain a list of jobs.

Elforama:
Currently users will have 0 to 10 messages most likely.

Good catch on the networking call. I don't remember, but I pinged the server engineer who maintains it and cc'd you.

Elforama:
@sidsharma2002 the offset is just echoed back by the API, so the client controls it. If you wanted to make the API calls in parrallel that would make a lot of sense.

Elforama (approved):
Looks great, thanks for the improvements. Sorry about the delay here. I will go ahead and rebase/merge it.

Elforama:
Actually, it looks like it's failing a unit test.

sidsharma2002:
@Elforama all checks have passed now

PR #1025, opened 2022-10-10
Button UI Revamp

sidsharma2002:
https://user-images.githubusercontent.com/53833109/194798389-2fc917a7-c755-49ee-a871-0d14d427da1f.mp4

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
@@ -503,7 +503,7 @@ fun DepthSurface(
 ) {
     val offsetAnimation: Float by animateFloatAsState(
         targetValue = if (isPressed) 1f else 0f,
-        animationSpec = tween(durationMillis = 100)
+        animationSpec = tween(durationMillis = 80)
     )
 
     when (shape) {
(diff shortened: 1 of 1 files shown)


sidsharma2002:
@sebastiangaertner @Elforama how does this looks like?

sebastiangaertner:
@sidsharma2002 it will def be a great improvement for the users. I like it. Jonathan is this something that can be included in the next version already?

EngrTaofeek:
@sidsharma2002 Great this animation looks better. Also leave the corner radius as it is or increase it so the buttons don't have very sharp edges.

EngrTaofeek:
This the Figma file although it's been inactive for a while https://www.figma.com/file/hn67qtbMN1aP3yrXx69dP2/Android-UX-(v2---w%2F-messaging-1.0.5)?node-id=0%3A386

EngrTaofeek:
@sebastiangaertner I was thinking since we have more UI/UX volunteers it might be better to let them review generally also since the project itself needs a design review. If we could try to set up a call this week that would be great. Jacob was unavailable last week so we couldn't get much done.

sidsharma2002:
> @sidsharma2002 this animation looks great. Also leave the corner radius as it is or increase it so the buttons don't have very sharp edges.

got it! 15dp is looks good to me

Elforama:
I think as long as we add back the regular button depth and soft edges, this is better than what we had. @sebastiangaertner  any concerns here? Otherwise I'd like to get this change in.

sidsharma2002:
@Elforama when are we thinking to get this merged?

Elforama:
Sorry for the delay here.

I tested this out and noticed a couple things.
1. We should leave the edge radius at 30
2. This only affects square buttons, giving an inconsistent feel. Can we also apply this update to circular buttons?

With those fixes I can merge this.

Elforama:
This was handled in another PR: https://github.com/Greenstand/treetracker-android/pull/1037

PR #1030, opened 2022-10-31
No messages ui

### LINKED ISSUE #964: Need 'no messages' UI for when a grower doesn't have any messages yet

EngrTaofeek:
Adds screen display to show users when the messages are empty. Fixes #964 
Screenshot_20221031_194656


app/src/main/java/org/greenstand/android/TreeTracker/messages/individualmeassagelist/IndividualMessageListScreen.kt
```diff
@@ -125,5 +126,8 @@ fun IndividualMessageListScreen(
                 }
             }
         }
+        if (state.messages.isEmpty()) {
```
Elforama, line 129: This looks good overall. Let's make sure we don't run the list code when it's empty.

Elforama (changes requested):

app/src/main/java/org/greenstand/android/TreeTracker/messages/individualmeassagelist/IndividualMessageListScreen.kt
EngrTaofeek: > This looks good overall. Let's make sure we don't run the list code when it's empty.

Sorry I don't get this

Elforama: Let's not run the code to render a list of it's empty

Elforama: If list empty show empty state, else show list

EngrTaofeek: Oh I get you now.

Elforama (approved):
Looks great, thanks for the fix

Issue #997, opened 2022-08-02
"Recapture" session button and later opt in in notes dialogue 


sebastiangaertner:
In order to distinguish between first captures (planting) and re-captures (maintenance visit or replanting) the app should provide a simple selection on what the session is. Practically a planter would go out to either captures newly planted trees or to re-track. To provide better data quality this should be done on the ground. Currently organisations fill the notes field with a number such as "1" (for the first re-tracking) or even "2021" for the tracking in 2021. In order to get better data quality we should make this an actual field_data entry.

sebastiangaertner:
@gwynndp What is needed in terms of data structure in the "field_data.session" ?

PR #1031, opened 2022-11-02
Added custom spinner to select capture phase

EngrTaofeek:
Added a custom spinner to select if a captured is initial capture. This addresses #997 partly. The next thing is to add a field to the tree capture request so we can send the chosen option with each tree.
[capture.webm](https://user-images.githubusercontent.com/38661872/199499448-32926e61-c490-49ba-a501-3aff3e7f6d00.webm)
untitled

@sebastiangaertner @Kpoke


Elforama (changes requested):
Who designed this?

We discussed how this would work months ago, and it was agreed that we'd select the capture mode when setting up the capture session

EngrTaofeek:
> Who designed this?
> 
> We discussed how this would work months ago, and it was agreed that we'd select the capture mode when setting up the capture session

Sorry I wasn't aware of the discussion or I totally forgot. I just designed it, when Sebastian brought the issue to my notice recently. At what point should it be done after the wallet selection in org picker screen and should we still retain the spinner for selection?

Elforama:
We want give the user this option during the capture setup flow.

PR #1033, opened 2022-11-13
Portuguese support

### LINKED ISSUE #1032: Portuguese Support for App
Currently there is no way to change language to Portuguese or detect automatically.
1. Add Portuguese to language selection options use cases in the app.
2. Auto-detect android devices with Portuguese as default language and set up tree tracker android app in Portuguese on installation.

EngrTaofeek:
Added Portuguese language to let users change the app language to Portuguese. Fixes #1032

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
@@ -274,7 +274,7 @@ fun BoxScope.LanguageButton() {
         colors = AppButtonColors.ProgressGreen,
         modifier = Modifier
             .align(Alignment.Center)
-            .size(width = 100.dp, 60.dp),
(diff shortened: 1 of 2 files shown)


app/src/main/java/org/greenstand/android/TreeTracker/view/TextButton.kt
```diff
@@ -274,7 +274,7 @@ fun BoxScope.LanguageButton() {
         colors = AppButtonColors.ProgressGreen,
         modifier = Modifier
             .align(Alignment.Center)
-            .size(width = 100.dp, 60.dp),
+            .size(width = 120.dp, 60.dp),
```
Elforama, line 277: There may be other languages, so we need to find a way to handle them without increasing the size of the button.

Elforama (changes requested):

PR #1039, opened 2022-12-16
added start time to sessionRequest object

erinfolami:


app/src/main/java/org/greenstand/android/TreeTracker/api/models/requests/SessionRequest.kt
```diff
@@ -15,4 +16,6 @@ data class SessionRequest(
     val organization: String,
     @SerializedName("device_configuration_id")
     val deviceConfigId: String,
+    @SerializedName("start_time")
+    val startTime: Instant,
```
Elforama, line 20: We'll need to send a 'Long' here instead.

Elforama (changes requested):

Elforama:
looking good overall

app/src/main/java/org/greenstand/android/TreeTracker/models/SessionUploader.kt
```diff
@@ -24,8 +24,7 @@ class SessionUploader(
                 targetWallet = session.destinationWallet,
                 organization = session.organization ?: "",
                 deviceConfigId = dao.getDeviceConfigById(session.deviceConfigId!!)!!.uuid,
-                startTime = session.startTime
-            )
+                startTime = session.startTime.toString().toLong()            )
```
Elforama, line 27: Please keep formatting in mind and try to follow how the rest of the code is formatting.

Elforama (changes requested):
Thanks for putting in these changes, however this breaks the upload process. Please ensure that you always test your changes before submitting a final PR.

Once the upload process works again I'll be able to merge this.

Elforama (approved):
Thanks for making the change. Looks good.

PR #1044, opened 2023-01-04
App Database Testing

precious3173:


app/src/androidTest/java/org/greenstand/android/TreeTracker/database/AppDatabaseTest.kt
```diff
@@ -0,0 +1,53 @@
...
+import org.hamcrest.CoreMatchers.*;
+import org.junit.After
+import org.junit.Assert.*
+import org.junit.Before
+import org.junit.Test
+import org.junit.runner.RunWith
+import java.io.IOException
+
+@RunWith(AndroidJUnit4::class)
+class AppDatabaseTest: TestCase(){
```
Elforama, line 19: The TreTrackerDAO has had a test added to it already. Feel free to delete this and add any extra test cases to the already created test file.

app/src/main/java/org/greenstand/android/TreeTracker/database/entity/LocationEntity.kt
```diff
@@ -16,7 +16,7 @@ import androidx.room.PrimaryKey
         )
     ]
 )
-data class LocationEntity(
+data class  LocationEntity(
```
Elforama, line 19: There are a ton of files with small changes. Please avoid making changes to files that are unrelated to the goal of the PR. It makes it more likely that a merge conflict will happen.

app/build.gradle
```diff
@@ -266,8 +269,39 @@ dependencies {
...
+
+        // The following Espresso dependency can be either "implementation",
+        // or "androidTestImplementation", depending on whether you want the
+        // dependency to appear on your APK"s compile classpath or the test APK
+        // classpath.
+        androidTestImplementation( "androidx.test.espresso:espresso-idling-resource:3.5.0")
+
+        debugImplementation "androidx.fragment:fragment-testing:1.5.5"
+
+        debugImplementation "androidx.test:monitor:1.6.0"
```
Elforama, line 303: Please undo all of this. There is no need for it.

Elforama (changes requested):
Thanks for adding some tests. Please see my comments on how to move this PR forward.

app/build.gradle
precious3173: Ok noted. I will update as stated

PR #1047, opened 2023-01-05
add tests for capture and orgpicker viewmodels

Naphtali-cpu:
Added unit tests for the following view models:
**OrgPickerViewModel.kt**
**TreeCaptureViewModel.kt**
**TreeImageReviewViewModel.kt**

### DIFF
--- a/app/build.gradle
+++ b/app/build.gradle
@@ -224,6 +224,10 @@ dependencies {
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-core:1.5.0'
     implementation 'org.jetbrains.kotlinx:kotlinx-coroutines-android:1.5.0'
     implementation 'org.jetbrains.kotlinx:kotlinx-datetime:0.3.2'
+    testImplementation 'junit:junit:4.12'
+    testImplementation 'junit:junit:4.12'
+    testImplementation 'junit:junit:4.12'
+    testImplementation 'junit:junit:4.12'
     coreLibraryDesugaring("com.android.tools:desugar_jdk_libs:1.1.5")
 
     //Database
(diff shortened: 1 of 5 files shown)


app/src/test/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerViewModelTest.kt
```diff
@@ -0,0 +1,68 @@
...
+     Test if setOrg sets to the current org to null.
+     */
+
+    @Test
+    fun `test for set org`() = runBlockingTest {
+        val org = null
+        org?.let { orgPickerViewModel.setOrg(it) }
+        val actualState = orgPickerViewModel.state.getOrAwaitValueTest()
+        assertEquals(org, actualState.currentOrg)
+    }
```
Elforama, line 67: This test does not actually test anything.

To test the org viewmodel we should:
- ensure it loads the proper current org returned from the orgRepo, as well as the list all orgs. The current org should always be in the list of orgs
- verify the orgRepo gets a new org set when we use the viewmodel to set a new org

Elforama (changes requested):
Thanks for the fast turnaround on this. I left some comments. Once they are addressed I can get this merged.

app/build.gradle
```diff
@@ -262,6 +266,7 @@ dependencies {
     implementation 'com.google.firebase:firebase-iid:21.0.1'
 
     testImplementation "io.mockk:mockk:1.10.0"
+    testImplementation "org.mockito:mockito-inline:3.11.2"
```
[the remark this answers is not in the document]
Naphtali-cpu: Noted.

app/src/test/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerViewModelTest.kt
Naphtali-cpu: I am having a problem with this test specifically while implementing it as you have requested. I'll ignore this test for now as I continue to see how i'll go about it. Or if there is anyone willing to help it will be better :)

PR #1051, opened 2023-01-09
Migrate to flow

kanake10:
PR to use flow in our WalletSelectViewModel

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/walletselect/WalletSelectScreen.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/walletselect/WalletSelectScreen.kt
@@ -43,7 +43,7 @@ fun WalletSelectScreen(
     viewModel: WalletSelectViewModel = viewModel(factory = LocalViewModelFactory.current)
 ) {
 
-    val state = viewModel.state.collectAsState(initial = WalletSelectState())
+    val state by viewModel.state.collectAsState(initial = WalletSelectState())
 
     val navController = LocalNavHostController.current
     val scope = rememberCoroutineScope()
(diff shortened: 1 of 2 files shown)


Elforama (changes requested):
Thanks for submitting your first PR!

Your title says this is migrating to flow, but your actually migrating to compose mutableState. All the viewmodels in this app either use Flow or LiveData (ideally flow). Can you please change this to use Flow?

app/src/main/java/org/greenstand/android/TreeTracker/walletselect/WalletSelectScreen.kt
```diff
@@ -42,7 +43,7 @@ fun WalletSelectScreen(
     viewModel: WalletSelectViewModel = viewModel(factory = LocalViewModelFactory.current)
 ) {
 
-    val state by viewModel.state.observeAsState(initial = WalletSelectState())
+    val state = viewModel.state.collectAsState(initial = WalletSelectState())
```
Elforama, line 46: Use 'by' instead of '='

app/src/main/java/org/greenstand/android/TreeTracker/walletselect/WalletSelectViewModel.kt
```diff
@@ -32,10 +31,8 @@ class WalletSelectViewModel(private val userRepo: UserRepo) : ViewModel() {
                     _state.value = _state.value?.copy(
                         currentUser = currentUser,
```
Elforama, line 32: Now that it's Flow, I don't think we need the '?.' anymore

Elforama (changes requested):
Thanks for the changes. Just some minor changes and I can merge it.

app/src/main/java/org/greenstand/android/TreeTracker/walletselect/WalletSelectViewModel.kt
kanake10: okay,
Going to replace this with a dot call

Elforama (approved):
Looks great, thanks!

PR #1054, opened 2023-01-11
Testing tree capture view model

ArtemPolovin:


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
 #Sat Apr 17 14:55:21 EDT 2021
 distributionBase=GRADLE_USER_HOME
-distributionUrl=https\://services.gradle.org/distributions/gradle-7.0.2-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-7.4-bin.zip
```
Elforama, line 3: Please undo this. It's not needed for a unit test.

build.gradle
```diff
@@ -9,11 +9,11 @@ buildscript {
...
 
     dependencies {
-        classpath 'com.android.tools.build:gradle:7.0.0'
-        classpath 'com.google.gms:google-services:4.3.3'
+        classpath 'com.android.tools.build:gradle:7.3.0'
+        classpath 'com.google.gms:google-services:4.3.10'
         classpath 'com.google.firebase:firebase-crashlytics-gradle:2.5.2'
         classpath "org.jetbrains.kotlin:kotlin-gradle-plugin:$kotlin_version"
-        classpath "androidx.navigation:navigation-safe-args-gradle-plugin:2.3.5"
+        classpath 'androidx.navigation:navigation-safe-args-gradle-plugin:2.4.1'
```
Elforama, line 16: Please undo this, it's not needed for a unit test.

app/build.gradle
```diff
@@ -72,12 +72,6 @@ android {
         coreLibraryDesugaringEnabled = true
     }
 
-    lintOptions {
```
Elforama, line 75: Please leave this gradle file alone unless making a PR explicitly to edit it, or if it's needed for the code changes (in this case it's not needed)

app/src/test/java/org/greenstand/android/TreeTracker/capture/TreeCaptureViewModelTest.kt
```diff
@@ -0,0 +1,124 @@
...
+            userRepo = userRepo,
+            treeCapturer = treeCapturer,
+            sessionTracker = sessionTracker,
+            createFakeTreesUseCase = createFakeTreesUseCase,
+            locationDataCapturer = locationDataCapturer
+        )
+    }
+
+    @Test
+    fun isFirstTrack() = runBlocking {
```
Elforama, line 68: Please use the `` syntax and describe in a bit more detail what the method is testing. For example:

fun `WHEN tree count greater than 0 THEN firstTrack is true`() {
    ... your code here
}

Also, this tests when it's greater than one, but this means we should also have a test confirming it's true if the tree count is less than 1.

app/src/main/AndroidManifest.xml
```diff
@@ -1,6 +1,5 @@
 <?xml version="1.0" encoding="utf-8"?>
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="org.greenstand.android.TreeTracker"
```
Elforama, line 3: Why did you remove this?

Elforama (changes requested):
Looking good! Please address the comments and rebase onto the latest changes that master has and I can merge this.

Elforama (approved):
Thanks for the cleanup and fixes. This looks great!

PR #1055, opened 2023-01-12
OrgPickerViewModelTest

Mzazi25:
Unit tests for OrgPickerViewModel Test

### DIFF
--- a/app/src/test/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerViewModelTest.kt
+++ b/app/src/test/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerViewModelTest.kt
@@ -3,6 +3,7 @@ package org.greenstand.android.TreeTracker.orgpicker
 import androidx.arch.core.executor.testing.InstantTaskExecutorRule
 import io.mockk.*
 import kotlinx.coroutines.ExperimentalCoroutinesApi
+import kotlinx.coroutines.runBlocking
 import kotlinx.coroutines.test.runBlockingTest
 import org.greenstand.android.TreeTracker.MainCoroutineRule
 import org.greenstand.android.TreeTracker.models.organization.OrgRepo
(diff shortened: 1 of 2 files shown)


Elforama (changes requested):
Looks good overall. Can we also verify that the orgRepo gets the proper org set?

Mzazi25:
I made the changes, kindly review

app/src/test/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerViewModelTest.kt
```diff
@@ -0,0 +1,51 @@
...
+    private lateinit var orgPickerViewModel: OrgPickerViewModel
+
+    @Before
+    fun setup(){
+        orgPickerViewModel = OrgPickerViewModel(orgRepo)
+    }
+
+    @Test
+    @Throws(Exception::class)
+    fun `set fake organization, returns success with correct data`()= runBlockingTest {
```
Elforama, line 35: We should break things down into two test methods.
1. Test that the state instantly gets the latests org from the repo. This tests the logic in the `init` block.
2. Test that when we set a new org, the selected org in the state updates and the repo assigns that new org to itself. This tests the logic in the setOrg method.

Elforama (changes requested):

app/src/test/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerViewModelTest.kt
Mzazi25: made adjustments.

Elforama (approved):

app/src/test/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerViewModelTest.kt
```diff
@@ -0,0 +1,59 @@
...
+
+    @Before
+    fun setup(){
+        orgPickerViewModel = OrgPickerViewModel(orgRepo)
+    }
+
+
+    @Test
+    @Throws(Exception::class)
+    fun `verify Org Repo gets correct Org Set`() = runBlockingTest{
```
Elforama, line 37: Ideally this would assert that the state has an initial state set by the init method, not just that the repo was called.

app/src/test/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerViewModelTest.kt
```diff
@@ -0,0 +1,59 @@
...
+        coEvery { orgRepo.getOrgs() } returns orgList
+
+        coVerify { orgRepo.getOrgs() }
+    }
+
+    @Test
+    @Throws(Exception::class)
+    fun `set fake organization, returns success with valid Org`()= runBlocking {
+        val currentOrg = FakeFileGenerator.fakeOrganizationList.first()
+        //Given
```
Elforama, line 48: No need for comments like these

app/src/test/java/org/greenstand/android/TreeTracker/utils/FakeFileGenerator.kt
```diff
@@ -191,4 +193,57 @@ object FakeFileGenerator {
         value = "random value",
         treeCaptureId = 1212
     )
+
+    val fakeOrganizationList = listOf(
```
Elforama, line 197: No need to call this "fake" since it's in a file titled "Fake"

app/src/test/java/org/greenstand/android/TreeTracker/orgpicker/OrgPickerViewModelTest.kt
```diff
@@ -0,0 +1,59 @@
...
+    @Throws(Exception::class)
+    fun `verify Org Repo gets correct Org Set`() = runBlockingTest{
+        val orgList = FakeFileGenerator.fakeOrganizationList
+        coEvery { orgRepo.getOrgs() } returns orgList
+
+        coVerify { orgRepo.getOrgs() }
+    }
+
+    @Test
+    @Throws(Exception::class)
```
Elforama, line 45: Neither of these methods throw an exception, so there is no need for this annotation.

PR #1064, opened 2023-01-21
IndividualMessageListViewModel Test

Mzazi25:
Added unit tests for IndividualMessageList Viewmodel

### DIFF
--- a/app/src/test/java/org/greenstand/android/TreeTracker/messages/individualmessagelist/IndividualMessageListViewModelTest.kt
+++ b/app/src/test/java/org/greenstand/android/TreeTracker/messages/individualmessagelist/IndividualMessageListViewModelTest.kt
@@ -1,0 +1,31 @@
+package org.greenstand.android.TreeTracker.messages.individualmessagelist
+
+import androidx.arch.core.executor.testing.InstantTaskExecutorRule
+import io.mockk.mockk
+import kotlinx.coroutines.ExperimentalCoroutinesApi
+import org.greenstand.android.TreeTracker.MainCoroutineRule
+import org.greenstand.android.TreeTracker.messages.individualmeassagelist.IndividualMessageListViewModel
+import org.greenstand.android.TreeTracker.models.UserRepo
+import org.greenstand.android.TreeTracker.models.messages.MessagesRepo
+import org.junit.Before
+import org.junit.Rule
+
+@ExperimentalCoroutinesApi
+class IndividualMessageListViewModelTest {
+    @get:Rule
+    var instantTaskExecutorRule = InstantTaskExecutorRule()
+
+    @get:Rule
+    var mainCoroutineRule = MainCoroutineRule()
+    private val userId = -33234254353456
+    private val userRepo = mockk<UserRepo>(relaxed = true)
+    private val messagesRepo = mockk<MessagesRepo>(relaxed = true)
+    private lateinit var individualMessageListViewModel: IndividualMessageListViewModel
+
+    @Before
+    fun setup(){
+        individualMessageListViewModel = IndividualMessageListViewModel(userId, userRepo, messagesRepo)
+    }
+
+
+}
--- a/app/src/test/java/org/greenstand/android/TreeTracker/messages/individualmessagelist/IndividualMessageListViewModelTest.kt
+++ b/app/src/test/java/org/greenstand/android/TreeTracker/messages/individualmessagelist/IndividualMessageListViewModelTest.kt
@@ -28,47 +28,31 @@ class IndividualMessageListViewModelTest {
     private val userId = 2L
     private val userRepo = mockk<UserRepo>(relaxed = true)
     private val messagesRepo = mockk<MessagesRepo>(relaxed = true)
-    private lateinit var individualMessageListViewModel: IndividualMessageListViewModel
+    private lateinit var testSubject: IndividualMessageListViewModel
 
     @Before
     fun setup(){
         coEvery { userRepo.getUser(any()) } returns FakeFileGenerator.fakeUsers.first()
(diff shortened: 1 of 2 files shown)


app/src/test/java/org/greenstand/android/TreeTracker/messages/individualmessagelist/IndividualMessageListViewModelTest.kt
```diff
@@ -0,0 +1,74 @@
...
+class IndividualMessageListViewModelTest {
+    @get:Rule
+    var instantTaskExecutorRule = InstantTaskExecutorRule()
+
+    @get:Rule
+    var mainCoroutineRule = MainCoroutineRule()
+    private val userId = 2L
+    private val userRepo = mockk<UserRepo>(relaxed = true)
+    private val messagesRepo = mockk<MessagesRepo>(relaxed = true)
+    private lateinit var individualMessageListViewModel: IndividualMessageListViewModel
```
Elforama, line 31: This is just preference, but I think naming whatever class you're testing as "testSubject" makes it very clear what's being tested. That allows you to look at any test and instantly make out what is being tested.

app/src/test/java/org/greenstand/android/TreeTracker/messages/individualmessagelist/IndividualMessageListViewModelTest.kt
```diff
@@ -0,0 +1,74 @@
...
+
+    @Before
+    fun setup(){
+        coEvery { userRepo.getUser(any()) } returns FakeFileGenerator.fakeUsers.first()
+        coEvery { messagesRepo.getMessageFlow(any()) } returns flowOf(FakeFileGenerator.messages)
+        individualMessageListViewModel = IndividualMessageListViewModel(userId, userRepo, messagesRepo)
+    }
+
+    @Test
+    @Throws(Exception::class)
```
Elforama, line 41: Unless you expect this to throw an exception, you don't need to add these to every method.

app/src/test/java/org/greenstand/android/TreeTracker/messages/individualmessagelist/IndividualMessageListViewModelTest.kt
```diff
@@ -0,0 +1,74 @@
...
+       coVerify { userRepo.getUser(userId) }
+    }
+
+    @Test
+    @Throws(Exception::class)
+    fun `verify message repo gets the correct message flow`() = runBlocking {
+        val currentUser = userRepo.getUser(userId)
+        val wallet = currentUser!!.wallet
+        coVerify { messagesRepo.getMessageFlow(wallet) }
+    }
```
Elforama, line 52: This method does not test the testSubject, and therefor should not be here.

Instead we just want to setup the mocked data and then check that the testSubjects state is correct.

app/src/test/java/org/greenstand/android/TreeTracker/messages/individualmessagelist/IndividualMessageListViewModelTest.kt
```diff
@@ -0,0 +1,74 @@
...
+    @Throws(Exception::class)
+    fun `WHEN message flow is triggered THEN message state updates with correct message`()= runBlocking {
+        val result = individualMessageListViewModel.state.getOrAwaitValueTest().messages.first()
+        assertEquals(result, FakeFileGenerator.messages.first())
+    }
+    @Test
+    @Throws(Exception::class)
+    fun `WHEN message flow is triggered THEN current user state updates with correct User`()= runBlocking {
+        val result = individualMessageListViewModel.state.getOrAwaitValueTest().currentUser
+        assertEquals(result, FakeFileGenerator.fakeUsers.first())
```
Elforama, line 64: I think these can be one test. You're verifying the state object when the testSubject is first created.

Elforama (changes requested):

app/src/test/java/org/greenstand/android/TreeTracker/messages/individualmessagelist/IndividualMessageListViewModelTest.kt
```diff
@@ -0,0 +1,58 @@
...
+
+    @Test
+    fun `verify user repo gets the correct user`() = runBlocking {
+       coVerify { userRepo.getUser(userId) }
+    }
+
+    @Test
+    fun `WHEN message flow is triggered THEN state updates with correct message and user`()= runBlocking {
+        val message = testSubject.state.getOrAwaitValueTest().messages.first()
+        val currentUser = testSubject.state.getOrAwaitValueTest().currentUser
```
Elforama, line 48: This is okay. But moving forward I'd suggest putting the state into a local variable instead of accessing it from the viewmodel each time. There could be a case in the future where this returns a different result breaking this test.

app/src/test/java/org/greenstand/android/TreeTracker/messages/individualmessagelist/IndividualMessageListViewModelTest.kt
```diff
@@ -0,0 +1,58 @@
...
+        assertEquals(message, FakeFileGenerator.messages.first())
+        assertEquals(currentUser, FakeFileGenerator.fakeUsers.first())
+    }
+    @Test
+    fun `WHEN selected message is triggered THEN state updates with correct message`()= runBlocking {
+        testSubject.selectMessage(FakeFileGenerator.fakeSurveyMessage)
+        val result = testSubject.state.getOrAwaitValueTest().selectedMessage
+        assertEquals(result, FakeFileGenerator.messages[2])
+    }
+}
```
Elforama, line 58: I should have mentioned this before. I think it would be useful to also verify that the selected message is null, then once we select a message, we verify it's no longer null.

Elforama (approved):

PR #1067, opened 2023-01-24
Run lint in CI

kanake10:
get linting to work in the CI


Elforama:
We should try integrating this into the current PR workflow that we have, that should also resolve the error that this PR is getting.

PR #1077, opened 2023-02-23
Test splash screen view model

ArtemPolovin:


app/src/test/java/org/greenstand/android/TreeTracker/splash/SplashScreenViewModelTest.kt
```diff
@@ -0,0 +1,144 @@
...
+            locationDataCapturer = locationDataCapturer,
+            messagesRepo = messagesRepo,
+            checkForInternetUseCase = checkForInternetUseCase,
+            orgRepo = orgRepo,
+            exceptionDataCollector = exceptionDataCollector
+        )
+    }
+
+    @Test
+    fun `all functions are executed when condition is true`() = runBlocking {
```
Elforama, line 82: Can you please update all the function names to be more descriptive about what they are testing?

Elforama (changes requested):
Thanks for adding this. It's looking good, just have a minor comment.

app/src/test/java/org/greenstand/android/TreeTracker/splash/SplashScreenViewModelTest.kt
ArtemPolovin: Fixed

Elforama (approved):

Issue #1057, opened 2023-01-13
Tree Upload Reminder Dialog


Elforama:
There can be times that a planter is planting tons of trees but hasn't uploaded anything. This dialog should inform the user that they should upload their trees soon.

Trigger dialog when clicking the "Track" button.
Dialog should show if non-uploaded trees is greater than 2000.

Dialog Title: "Upload Trees Soon"
Dialog Body: "You have over 2000 trees, please upload as soon as you can."
Dialog Button options: "Okay" (this should dismiss the dialog)

PR #1081, opened 2023-03-17
Tree Upload Reminder Dialog

Mzazi25:
This PR adds Alert Dialog to users who have planted trees above 2000 and reminds them to upload the trees. 
Fixes on Issue #1057


app/src/main/java/org/greenstand/android/TreeTracker/dashboard/DashboardScreen.kt
```diff
@@ -63,13 +61,30 @@ fun DashboardScreen(
     val state by viewModel.state.observeAsState(DashboardState())
     val snackBar by viewModel.snackBar.observeAsState()
     val navController = LocalNavHostController.current
+    var dialogOpened by remember { mutableStateOf(false) }
+    var showDialog by remember { mutableStateOf(false) }
 
+    if (showDialog){
+        DisplayAlertDialog(
+            title = "Upload Trees Soon",
+            message ="You have over 2000 trees, please upload as soon as you can.",
```
Elforama, line 70: Text that the user sees should always use string resources to allow translations.

app/src/main/java/org/greenstand/android/TreeTracker/dashboard/DashboardScreen.kt
```diff
@@ -63,13 +61,30 @@ fun DashboardScreen(
...
+        )
+    }
     Dashboard(
         state = state,
         snackBar = snackBar,
         onSyncClicked = { viewModel.sync() },
         onOrgClicked = { navController.navigate(NavRoute.Org.route) },
-        onCaptureClicked = { navController.navigate(NavRoute.UserSelect.route) },
+        onCaptureClicked = {
+            if (state.totalTreesToSync >=2000){
```
Elforama, line 84: We can move this check into a boolean method on the state object that returns true or false

app/src/main/java/org/greenstand/android/TreeTracker/dashboard/components/DisplayAlertDialog.kt
```diff
@@ -0,0 +1,48 @@
...
+
+import androidx.compose.material.AlertDialog
+import androidx.compose.material.Button
+import androidx.compose.material.Text
+import androidx.compose.runtime.Composable
+import androidx.compose.ui.text.font.FontWeight
+import androidx.compose.ui.unit.sp
+
+@Composable
+fun DisplayAlertDialog(
```
Elforama, line 11: We have a dialog method we can reuse that has a special style that matches the app. Please use that instead

Elforama (changes requested):
Thanks for tackling this. There are some required changes, but it looks good overall.

Please keep formatting in mind while working, try to match the style in the file.

app/src/main/java/org/greenstand/android/TreeTracker/dashboard/DashboardScreen.kt
```diff
@@ -63,13 +61,29 @@ fun DashboardScreen(
     val state by viewModel.state.observeAsState(DashboardState())
     val snackBar by viewModel.snackBar.observeAsState()
     val navController = LocalNavHostController.current
+    var dialogOpened by remember { mutableStateOf(false) }
```
Elforama, line 64: dialogOpened doesn't seem to be used

app/src/main/java/org/greenstand/android/TreeTracker/dashboard/DashboardViewModel.kt
```diff
@@ -36,6 +36,7 @@ data class DashboardState(
     val totalTreesToSync: Int = 0,
     val isOrgButtonEnabled: Boolean = false,
     val showUnreadMessageNotification: Boolean = false,
+    val totalTreesToSyncThreshold:Boolean = false
```
Elforama, line 39: This should be renamed to something more descriptive like "showTreeSyncReminderDialog"

Elforama (changes requested):
Thanks for making the changes, it's looking a lot better. Can you please cleanup the last few things before I merge it?

app/src/main/java/org/greenstand/android/TreeTracker/dashboard/DashboardScreen.kt
Mzazi25: fixed

app/src/main/java/org/greenstand/android/TreeTracker/dashboard/DashboardViewModel.kt
Mzazi25: fixed

Elforama (approved):

PR #1090, opened 2023-04-23
Add config screen

Elforama:
Adds a debug screen to turn on/off different features for testing and release. This will later hook up to RemoteConfig from Firebase.

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/devoptions/DevOptionsScreen.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/devoptions/DevOptionsScreen.kt
@@ -1,0 +1,85 @@
+package org.greenstand.android.TreeTracker.devoptions
+
+import androidx.compose.foundation.layout.padding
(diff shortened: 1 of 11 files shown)


EngrTaofeek:
I love this idea, really a great way to compare different features while testing. I tried tracking the trees did the first three without forcing image size and the last three while forcing image size. I noticed all six were in same resolution(1088 x 1920) and about same size (177.6kb to 191.9kb) See screenshot below kindly ignore the first image it's the profile picture from the front camera.
Screenshot 2023-04-24 at 18 48 41
A possible cause might be the image capture size already been specified in the Camera.kt file this feature might not be able to override that. Also might be this PR isn't yet forcing the size of images.

Elforama:
This doesn't force the size yet. It just adds the dev screen

EngrTaofeek:
Oh if that's the case it's all good  just a few UI changes and addition . The bottom left arrow button, a header for the screen and the checked state color of the checkbox should be the primary colour(green) instead of blue.

PR #1097, opened 2023-05-29
Encrypt databases

S1210:
DB encryption #1058


EngrTaofeek:
@S1210 Thanks on your first PR, here are few notes:
1 The PR contains too many changes which makes it very tough to review. For instance there are some parts I would like to approve and some I won't.
2 Let each pull request address an issue and not multiple except if they are really intertwined and can't be done seperately.
3. This PR failed CI/CD test which is a requirement for pull request to be merged
4. Refactor should be done to improve code quality and not just renaming files or doing same thing in a different way with no direct improvement.
5. Library updates in grade file should also be tested very well as some library versions are not compatible with other versions

PR #1142, opened 2025-04-23
feat: view privacy policy from settings

### LINKED ISSUE #1141: Show Privacy Dialog from settings
Show Greenstand's Privacy Policy from settings screen as well.

EngrTaofeek:
Gives users the option to see privacy dialog from their settings as well and not only when they create a new account.


Davidezrajay:
Looks good on 2.2.1 (build 202)

PR #1161, opened 2025-07-04
Fix for issue #1157

### LINKED ISSUE #1157: Investigate Failing Unit tests
Currently all unit tests are failing, investigate reason why they are failing and possible fix.
HINT: It is likely they are failing due to updated dependencies.

Sakya-Kanuparthy:
Added note under "Project Setup" about Robolectric test failures with JDK 21.
Recommend using OpenJDK 17.0.15 LTS.

Fixes #1157

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,15 @@ Developers will need to ask the #android_chat channel in Slack for the treetrack
 
 For development, select the build variant _dev_. This build variant is configured to allow trees to be added without a specific accuracy.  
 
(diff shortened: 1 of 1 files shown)


EngrTaofeek:
Great spot on there @Sakya-Kanuparthy but instead of downgrading the JDK 21 to JDK 17 can't we update the RoboElectric dependency to a more recent one that would support JDK 21

Tarek-Bohdima:
fixed by #1179

PR #1255, opened 2026-04-05
chore: remove redundant Dispatchers.Main from viewModelScope.launch

### LINKED ISSUE #1232: Remove unnecessary Dispatchers.Main on viewModelScope.launch
## Problem

Several ViewModels specify `Dispatchers.Main` when launching coroutines on `viewModelScope`, which already defaults to `Dispatchers.Main`. This is misleading and masks the intent — it suggests the dispatcher choice was deliberate when it's actually redundant.

```kotlin
// Current (redundant):
viewModelScope.launch(Dispatchers.Main) { ... }

// Should be:
viewModelScope.launch { ... }
```

## Locations

- `app/src/main/java/org/greenstand/android/TreeTracker/capture/TreeCaptureViewModel.kt`
- `app/src/main/java/org/greenstand/android/TreeTracker/signup/SignupViewModel.kt`

## Suggested Fix

Remove the explicit `Dispatchers.Main` parameter from `viewModelScope.launch()` calls in these files. If IO work is needed, use `withContext(Dispatchers.IO)` inside the launch block instead.

## Effort
Low

evanofficial:
`viewModelScope.launch` already defaults to `Dispatchers.Main`, so passing it explicitly is redundant and misleads intent.

Fixes the two locations called out in the issue:

- `TreeCaptureViewModel.kt`
- `SignupViewModel.kt`

Also included a third matching occurrence in `TreeImageReviewViewModel.kt` (init block, same pattern) — I asked in the issue comment whether to roll it in. Happy to drop it from this PR if you'd rather keep scope tight to the two files listed.

Removed now-unused `kotlinx.coroutines.Dispatchers` imports in each of the three files.

Closes #1232


Elforama:
Thanks for the PR. Please confirm the app works with an attached video by testing the tree capture flow.

Elforama:
Thanks for the contribution! Please update it with the base branch

PR #1274, opened 2026-04-21
Fix white screen crash from rapid taps on language picker arrow

### LINKED ISSUE #1265: App crashes when repeatedly tapping bottom-right arrow on language selection screen.
### App Version / Build Number

2.2.1

### Device

ASUS ROG Phone 9

### Android Version

Android 16

### Describe the bug

The app crashes and displays a blank white screen when the user repeatedly taps the bottom-right arrow on the language selection screen.
This issue occurs when the user taps very quickly during screen transitions, causing the app to become unresponsive.

### To Reproduce

1. Open the app and stay on the main screen
2. Tap the "English" button on the top-right.
3. The app navigates to the language selection screen.
4. Select any language (English, Swahili, or Portuguese).
5. Immediately and rapidly tap the bottom-right arrow multiple times without waiting for the screen transition to complete.
6. Continue tapping even as the app starts navigating back to the main screen.
7. Observe the app behavior.

### Expected behavior

The app should handle multiple rapid taps without crashing and should navigate back to the main screen normally.

### Actual behavior

The app shows a white screen and becomes unresponsive, appearing to crash.

### Frequency

Always reproducible

### Severity

Blocker - App crashes or data loss

### Network Conditions

Wi-Fi

### Screenshots / Screen recordings

https://github.com/user-attachments/assets/ca6e2b2a-fc58-441d-ae04-9f7c247f1f6b

### Logs

```shell
No logs available.
```

### Additional context

This issue occurs when the user taps rapidly during screen transition, which may indicate a race condition or unhandled multiple input events.
The issue does not occur when tapping normally (slow interaction).
Tested on Android device with Wi-Fi connection.

brundamarpadaga:
Prevent multiple navController calls during screen transition by adding a hasNavigated guard flag in LanguageSelectScreen. Also adds two additional test cases to LanguagePickerViewModelTest.

Fixes #1265

Thank you for opening a Pull Request!

- [ ] Make sure to open a GitHub issue as a bug/feature request before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
- [x ] Ensure the tests are added/updated (if necessary)
- [ ] Ensure the linter passes (`./codeAnalysis` to automatically apply formatting/linting)
- [ ] Appropriate docs were updated (if necessary)

Fixes #1265  🦕


app/src/main/java/org/greenstand/android/TreeTracker/languagepicker/LanguageSelectScreen.kt
```diff
@@ -51,12 +54,15 @@ fun LanguageSelectScreen(
 ) {
     val state by viewModel.state.collectAsState()
     val navController = LocalNavHostController.current
+    var hasNavigated by remember { mutableStateOf(false) }
```
Elforama: Is is possible that the user navigates back to this screen and then they can't navigate forward anymore?

Also, lets move any logic like this into the viewmodel instead.

Elforama (changes requested):

app/src/main/java/org/greenstand/android/TreeTracker/languagepicker/LanguageSelectScreen.kt
brundamarpadaga: Hi @Elforama, thanks for the feedback! Back navigation was already handled — `remember` resets when the composable re-enters composition. I've now moved the logic into the ViewModel as suggested, using `hasNavigated`, `tryNavigate()`, and `ResetNavigation`. New push is up, let me know if you'd like any further changes!

Elforama:
Thanks for the changes.

I was testing this out and it seems to work, but noticed that this is an issue with more than just the language button. It's possible to rapid tap many navigation back button, and it will result in the app going white.

We should figure out a way to solve this for all screens. We should not use this same flag technique for all screens.

What do you think the reason is for the app going white? Any ideas on how to solve this for all screens?

brundamarpadaga:
Hi @Elforama , good point! The root cause is that Jetpack Compose's NavController doesn't prevent re-entrant navigation calls — if the user taps rapidly, multiple popBackStack() or navigate() calls fire before the first transition completes, leaving the back stack in an inconsistent state which causes the white screen.
A clean way to solve this globally without repeating the flag pattern on every screen would be to create a single extension function on NavController that guards against re-entrant calls.
Or more robustly, wrap popBackStack() and navigate() with a debounce at the NavController level so any rapid calls within a short window are dropped automatically. This way the fix lives in one place and all screens benefit from it without any per-screen changes.
Would you like me to implement this approach and update the PR accordingly?

Elforama (commented):
Thanks for the PR! Looking good so far. Can we name these methods throttledXXXX instead? 

Also, can you please attach a video showing the bug fixed?

Elforama (approved):
Good work, thank you!

PR #1277, opened 2026-05-05
BDD for treetracker app

dadiorchen:
This PR use BDD to describe and test the treetracker app, inlucding check the capture shows up in admin panel (upload succedded)

The feature is described by below:

https://github.com/Greenstand/treetracker-android/blob/3e5e5f7a538d6705f39f8de43447a73fb6a52ae4/e2e/features/02_signup_flow.feature#L1-L21

https://github.com/Greenstand/treetracker-android/blob/3e5e5f7a538d6705f39f8de43447a73fb6a52ae4/e2e/features/03_capture_setup.feature#L1-L23

And the video records all those steps: 
https://github.com/Greenstand/treetracker-android/blob/feat/bdd-for-2.1.3/e2e/test-artifacts/videos/feature-02-03-admin-panel.mp4

@Elforama please help me review this, the code change to the Kotlin is bascially adding accecibility text to support e2e. I think this is verfy promising to adopt to whole platofrom, with the Claude I don't think it is dfficult for mobile dev to write e2e in this way ( I didn't code specific lines)


Elforama (changes requested):
Let's use an android testing framework instead. Such as espresso or maestro.

Elforama:
If we want to test the full stack then we can move this code outside the android repo possibly. Also, have you looked into maestro? It allows writing tests that look like your BDD files instead of tons of JS code.

PR #1278, opened 2026-05-05
add BDD to 2.2.1

dadiorchen:
check the PR #1277 for description


Elforama (changes requested):
should use android testing library

PR #1279, opened 2026-05-06
feat:Hook firebase remote config org fetch into sync and smart startup

### LINKED ISSUE #1262: Hook Firebase Remote Config org fetch into sync and smart startup
## Summary

Firebase Remote Config is now used to fetch org configurations when an org deeplink is opened. However, the fetch only happens at deeplink time. If an org's config changes server-side, the app has no way to pick up the update until the user opens the deeplink again.

This issue adds two behaviors:
1. **Fetch org config during the main data sync** — so config updates are picked up regularly.
2. **Fetch org config on app startup, but only until the first successful fetch** — so the app gets config quickly after a deeplink, then stops wasting time on future cold starts.

## Current Behavior

- `OrgConfigProvider.fetchOrgConfig(orgId)` is called only in `SplashScreenViewModel.bootstrap()` when a deeplink provides an `orgId`.
- The main sync flow (`SyncDataUseCase.execute()` → `TreeSyncWorker`) does not touch Remote Config at all.
- On a normal app launch (no deeplink), Remote Config is never fetched, even if the current org was set by a previous deeplink.

### 1. Fetch during sync

When the user taps "Upload" on the dashboard and `SyncDataUseCase.execute()` runs, it should also refresh the current org's config from Remote Config.

**Where to add it:** `SyncDataUseCase.execute()` in `app/src/main/java/org/greenstand/android/TreeTracker/usecases/SyncDataUseCase.kt`.

**Implementation:**
- Add `OrgConfigProvider` and `OrgRepo` as constructor dependencies (injected via Koin in `AppModule.kt`).
- After the existing `messagesRepo.syncMessages()` call (around line 60), add a step:
  ```kotlin
  val currentOrg = orgRepo.currentOrg()
  if (currentOrg.id != OrgRepo.DEFAULT_ORG_ID) {
      val configJson = orgConfigProvider.fetchOrgConfig(currentOrg.id)
      if (configJson != null) {
          orgRepo.addOrgFromRemoteConfig(currentOrg.id, currentOrg.name, configJson)
      }
  }
  ```
- Skip the fetch if the current org is the default Greenstand org (ID `"-1"`), since that one has no Remote Config entry.
- If the fetch fails or times out, just continue with the rest of the sync — don't block tree uploads.

**DI change:** Update `AppModule.kt` where `SyncDataUseCase` is constructed (currently `factory { SyncDataUseCase(get(), get(), get(), get(), get(), get(), get(), get()) }`) to include the two new dependencies.

### 2. Fetch on startup (until first success)

On app startup, if the app has a non-default current org, fetch its config from Remote Config. Once the fetch succeeds once, record that in SharedPreferences and stop doing it on future startups.

**Where to add it:** `SplashScreenViewModel.bootstrap()` in `app/src/main/java/org/greenstand/android/TreeTracker/splash/SplashScreenViewModel.kt`.

**Implementation:**
- After `orgRepo.init()` and the existing deeplink handling block, add a new block for non-deeplink startup refresh:
  ```kotlin
  // Existing deeplink handling...
  if (!orgId.isNullOrBlank()) {
      // ... existing deeplink code ...
  } else {
      // Non-deeplink launch: refresh current org config if we haven't yet
      val currentOrg = orgRepo.currentOrg()
      if (currentOrg.id != OrgRepo.DEFAULT_ORG_ID && !hasCompletedInitialOrgSync()) {
          val configJson = orgConfigProvider.fetchOrgConfig(currentOrg.id)
          if (configJson != null) {
              orgRepo.addOrgFromRemoteConfig(currentOrg.id, currentOrg.name, configJson)
              markInitialOrgSyncComplete()
          }
      }
  }
  ```

**Preference key:** Add a new preference key for tracking whether the initial sync has completed. Pattern follows existing usage in the codebase:
  ```kotlin
  private val ORG_CONFIG_SYNCED_KEY = PrefKeys.SYSTEM_SETTINGS + PrefKey("org-config-synced")
  ```

- `hasCompletedInitialOrgSync()` checks if the pref is `true`.
- `markInitialOrgSyncComplete()` sets the pref to `true`.
- This pref should be reset to `false` when a new deeplink is processed (in the deeplink handling block), so the next startup will re-fetch.

**Where to put the helpers:** Either directly on `SplashScreenViewModel` (private methods using the existing `Preferences` — note: would need to inject `Preferences`), or as methods on `OrgRepo` (which already has `Preferences` injected).

Putting them on `OrgRepo` is cleaner since it already owns org state:
  ```kotlin
  // In OrgRepo.kt
  fun hasCompletedInitialOrgSync(): Boolean =
      prefs.getBoolean(ORG_CONFIG_SYNCED_KEY, false)

  fun markInitialOrgSyncComplete() {
      prefs.edit().putBoolean(ORG_CONFIG_SYNCED_KEY, true).commit()
  }

  fun resetInitialOrgSync() {
      prefs.edit().putBoolean(ORG_CONFIG_SYNCED_KEY, false).commit()
  }
  ```

Call `resetInitialOrgSync()` inside `addOrgFromRemoteConfig()` and `addMinimalOrg()` when processing a deeplink, so the flag is cleared whenever a new org link arrives.

### 3. Logging

All new Remote Config fetch calls should use the `OrgLink` Timber tag (already used throughout the org link flow) so they can be filtered with `adb logcat -s OrgLink`.

## Files to Modify

| File | Change |
|------|--------|
| `app/.../usecases/SyncDataUseCase.kt` | Add `OrgConfigProvider` + `OrgRepo` deps, fetch current org config during sync |
| `app/.../di/AppModule.kt` | Add new deps to `SyncDataUseCase` factory |
| `app/.../splash/SplashScreenViewModel.kt` | Add startup fetch for non-deeplink launches (until first success) |
| `app/.../models/organization/OrgRepo.kt` | Add `hasCompletedInitialOrgSync()`, `markInitialOrgSyncComplete()`, `resetInitialOrgSync()` with new pref key |

## Testing

1. Set up a Remote Config key for a test org. Open the deeplink. Verify config is applied.
2. Change the Remote Config value in Firebase Console. Tap "Upload" on dashboard. Verify the org config updates after sync.
3. Kill and reopen the app (no deeplink). Verify it fetches config on startup (check `OrgLink` logs).
4. Kill and reopen the app again. Verify it does NOT fetch on startup (the flag is set).
5. Open a new deeplink. Kill and reopen. Verify it fetches again on startup (flag was reset).
6. Test offline: airplane mode, open app. Verify startup fetch fails gracefully and the app still works.

jaycenusa:
- [x] Make sure to open a GitHub issue as a bug/feature request before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
- [x] Ensure the tests are added/updated (if necessary)
- [x] Ensure the linter passes (`./codeAnalysis` to automatically apply formatting/linting)
- [x] Appropriate docs were updated (if necessary)

Fixes #1262 🦕

**Files change:** 
- app/.../di/AppModule.kt
- app/.../splash/SplashScreenViewModel.kt
- app/.../models/organization/OrgRepo.kt

**Unit test files:**
- app/.../test/../splash/SplashScreenViewModelTest.kt
- app/.../test/../usecases/SyncDataUseCaseTest.kt


app/src/main/java/org/greenstand/android/TreeTracker/models/organization/OrgRepo.kt
```diff
@@ -122,6 +135,10 @@ class OrgRepo(
             dao.insertOrg(validatedEntity)
             setOrg(validatedEntity.id)
             Timber.tag(ORG_LINK_TAG).i("Org '$orgName' ($orgId) loaded from Remote Config")
+            resetInitialOrgSync()
+            Timber.tag(ORG_LINK_TAG).i("Org config sync reset")
+            addMinimalOrg(orgId, orgName)
+            Timber.tag(ORG_LINK_TAG).i("Minimal org '$orgName' ($orgId) added with default flows")
```
Elforama: What is the purpose of this?

app/src/main/java/org/greenstand/android/TreeTracker/splash/SplashScreenViewModel.kt
```diff
@@ -81,6 +81,18 @@ class SplashScreenViewModel(
...
+            val currentOrg = orgRepo.currentOrg()
+            if (currentOrg.id != OrgRepo.DEFAULT_ORG_ID && !orgRepo.hasCompletedInitialOrgSync()) {
+                val configJson = orgConfigProvider.fetchOrgConfig(currentOrg.id)
+                Timber.tag(ORG_LINK_TAG).d("Fetching Remote Config for org ${currentOrg.id}")
+                if (configJson != null) {
+                    orgRepo.addOrgFromRemoteConfig(currentOrg.id, currentOrg.name, configJson)
+                    Timber.tag(ORG_LINK_TAG).i("Add Org From Remote Config ${currentOrg.id}")
+                    orgRepo.markInitialOrgSyncComplete()
+                    Timber.tag(ORG_LINK_TAG).i("Mark Initial Org Sync Complete ${currentOrg.id}")
+                }
```
Elforama, line 82: The issue attached states that we want to fetch config when we have an org link.

Elforama: Side note, this is a ton of calls to the org repo, this means most of this logic can be a single org repo function call.

app/src/main/java/org/greenstand/android/TreeTracker/usecases/SyncDataUseCase.kt
```diff
@@ -57,6 +61,15 @@ class SyncDataUseCase(
...
                     messagesRepo.syncMessages()
+                    val currentOrg = orgRepo.currentOrg()
+                    if (currentOrg.id != OrgRepo.DEFAULT_ORG_ID) {
+                        val configJson = orgConfigProvider.fetchOrgConfig(currentOrg.id)
+                        Timber.tag("OrgConfigProvider").d("Fetched config for org ${currentOrg.id}")
+                        if (configJson != null) {
+                            orgRepo.addOrgFromRemoteConfig(currentOrg.id, currentOrg.name, configJson)
+                            Timber.tag("OrgRepo").i("AddOrgFromRemoteConfig completed ${currentOrg.name}, ${currentOrg.id}, $configJson")
+                        }
+                    }
```
Elforama: This should be internal to org repo, and we should unconditionally fetch the config. And it should be outside other sync steps, and done async.

Elforama (changes requested):
Thanks for the PR! I left some feedback to be addressed.

app/src/main/java/org/greenstand/android/TreeTracker/models/organization/OrgRepo.kt
jaycenusa: The purpose of the code is to clear the ORG_CONFIG_SYNCED_KEY flag to false after processing a deeplink

Quoting from the issue: 

Image

app/src/main/java/org/greenstand/android/TreeTracker/splash/SplashScreenViewModel.kt
jaycenusa: You are right. I will revised with the new change.

app/src/main/java/org/greenstand/android/TreeTracker/models/organization/OrgRepo.kt
Elforama: That requirement was a mistake, you can remove this.

app/src/main/java/org/greenstand/android/TreeTracker/splash/SplashScreenViewModel.kt
```diff
@@ -72,7 +72,15 @@ class SplashScreenViewModel(
         if (!orgId.isNullOrBlank()) {
             Timber.tag(ORG_LINK_TAG).i("Deeplink received: orgId=$orgId, orgName=$orgName")
             val configJson = orgConfigProvider.fetchOrgConfig(orgId)
-            if (configJson != null) {
+            // Non-deeplink launch: refresh current org config if we haven't yet
```
Elforama, line 75: this is actually a deeplink launch.

Elforama (changes requested):
I've updated the issue to be more clear. I hope things make a bit more sense now.

So this needs to sync one time on startup, and every time when the sync process happens.

PR #1304, opened 2026-06-17
feat: Centralize shared keyboard options

thetwai-dev:
Thank you for opening a Pull Request!

- [X] Make sure to open a GitHub issue as a bug/feature request before writing your code! That way we can discuss the change, evaluate designs, and agree on the general idea
- [ ] Ensure the tests are added/updated (if necessary)
- [X] Ensure the linter passes (`./codeAnalysis` to automatically apply formatting/linting)
- [ ] Appropriate docs were updated (if necessary)

## 📸 Proof of change (REQUIRED)

> **Every PR must include a screen recording / video showing the change working.**
> **UI changes must also include before/after screenshots.**
> Just drag-and-drop the files into the boxes below — GitHub will upload them.
> PRs without the required media will be flagged automatically and **cannot be merged**.

### 🖼️ Screenshots (required for any UI change)

| Before | After |
| ------ | ----- |
|        |       |

- [X] This change has **no user-visible / UI effect** (refactor, docs, CI, etc.), so screenshots are not applicable.
  - A maintainer may also apply the `non-ui` label. **A video is still required** even for non-UI changes.
  
Addresses #736 partially. Only organize code, not hide suggestions.

### DIFF
--- a/app/src/main/java/org/greenstand/android/TreeTracker/messages/directmessages/ChatScreen.kt
+++ b/app/src/main/java/org/greenstand/android/TreeTracker/messages/directmessages/ChatScreen.kt
@@ -35,7 +35,6 @@ import androidx.compose.foundation.lazy.LazyListState
 import androidx.compose.foundation.lazy.rememberLazyListState
 import androidx.compose.foundation.shape.RoundedCornerShape
 import androidx.compose.foundation.text.KeyboardActions
-import androidx.compose.foundation.text.KeyboardOptions
 import androidx.compose.material.Scaffold
 import androidx.compose.material.Text
 import androidx.compose.material.TextField
@@ -50,7 +49,6 @@ import androidx.compose.ui.platform.testTag
 import androidx.compose.ui.res.painterResource
 import androidx.compose.ui.res.stringResource
 import androidx.compose.ui.text.font.FontWeight
-import androidx.compose.ui.text.input.ImeAction
 import androidx.compose.ui.text.input.KeyboardType
 import androidx.compose.ui.unit.dp
 import androidx.lifecycle.viewmodel.compose.viewModel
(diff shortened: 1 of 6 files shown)


Tarek-Bohdima (changes requested):
Thanks for picking up such an old one, @thetwai-dev — and for digging into the keyboard options. I ran both approaches on an emulator (Pixel 7, Gboard) to ground this. A couple of things before it can land:

**1. `KeyboardType.Password` — it works, with one tradeoff.**
On Gboard it's actually the *only* approach that hides the suggestion strip. I also tried the "proper" route — `InterceptPlatformTextInput` + `TYPE_TEXT_FLAG_NO_SUGGESTIONS` — and Gboard ignores it: the strip still shows even with autocorrect disabled. So your choice is the right one for Gboard. The one visible tradeoff is that Gboard switches to its password layout (number row, no emoji/glide keys). I'm deferring the final yes/no on the approach to @Elforama since he researched this previously.

**2. Scope — this is one field of several.** #736 asks to hide suggestions on *all* text inputs. These still use `KeyboardType.Text` with suggestions on:
- `ChatScreen.kt:170`
- `SessionNoteScreen.kt:104`
- `NameEntryView.kt:107` and `:134`
- `AddWalletScreen.kt:110`

Rather than copy-paste the same option into each, please centralize it — a single shared `KeyboardOptions` constant (or a small shared text-field composable) that all these fields use. Then change `Fixes #736` → `Addresses #736` so the issue stays open until every field is covered.

**3. Minor — API consistency.** You switched this one call site to `autoCorrectEnabled` while 7 others still use the deprecated `autoCorrect`. Both compile on Compose 1.7.5, but let's not mix spellings — migrate them together (ideally via the shared component above) or leave this one as `autoCorrect` for now.

No rush — happy to help once the approach is settled with Jonathan.

Elforama (approved):

PR #1308, opened 2026-06-27
test: verify require-pr-media check (DO NOT MERGE)

Elforama:
🚧 **Smoke test for the new `require-pr-media` workflow — please do not merge.**

Now updated to include media so we can confirm the check flips to ✅.

### 🎥 Screen recording
https://github.com/user-attachments/assets/00000000-1111-2222-3333-444444444444

### 🖼️ Screenshot
after

### DIFF
--- a/docs/tasks/README.md
+++ b/docs/tasks/README.md
@@ -1,0 +1,2 @@
+
+<!-- media-check smoke test: safe to delete this PR -->


Elforama:
Smoke test complete — check + comment behavior verified (fail→pass on media added). Closing.

